PR #19, opened 2020-02-21
build: fix pipcook-core lint errors

yorkie:
Escape is always good, so it disables the rule "no-useless-escape", too.


packages/pipcook-core/src/index.ts
```diff
@@ -1,7 +1,7 @@
 // types
 export {OriginSampleData, UniformSampleData, UniformTfSampleData, UniformGeneralSampleData} from './types/data';
 export {PipcookModel} from './types/model';
-export {DataDescriptor, metaData, EvaluateResult} from './types/other';
+export {DataDescriptor, MetaData, EvaluateResult} from './types/other';
```
yorkie: @utkobe Please help me check if I have missed anything about this, thank you :)

PR #32, opened 2020-03-03
plugin: move bayesian text classification to python bridge

utkobe:
The PR tries to move the bayesian text classification to python bridge way. One of the main reasons is enable it to deploy to EAS since currently EAS only supports python deployment. @yorkie  Please have a review


example/pipcook-app-example/pipcook-textclass-app-test.js
```diff
@@ -0,0 +1,5 @@
+const pipcook = require('@pipcook/pipcook-app').default;
+
+const classifier = new pipcook.textClassification('bayes')
+
+classifier.train('http://ai-sample.oss-cn-hangzhou.aliyuncs.com/image_classification/datasets/textDataBinding.csv', {}, false);
```
yorkie: Missing EOF.

example/pipeline-example/pipeline-text-bayes-classification.js
```diff
@@ -1,10 +1,9 @@
 const {DataCollect, DataAccess, DataProcess, ModelLoad, ModelTrain, ModelEvaluate, ModelDeploy, PipcookRunner} = require('@pipcook/pipcook-core');
 
 const textClassDataAccess = require('@pipcook/pipcook-plugins-text-csv-data-access').default;
-const textClassDataProcess = require('@pipcook/pipcook-plugins-text-class-data-process').default
 const bayesianClassiferModelLoader = require('@pipcook/pipcook-plugins-bayesian-classifier-model-load').default;
 const bayesianClassifierModelTrain = require('@pipcook/pipcook-plugins-bayesian-classifier-model-train').default;
-const classModelEvalute = require('@pipcook/pipcook-plugins-model-evaluate').default;
+const classModelEvalute = require('@pipcook/pipcook-plugins-bayesian-classifier-model-evaluate').default;
```
yorkie: How about using ES6 modules here?

packages/pipcook-app/src/models/textClassification.ts
```diff
@@ -0,0 +1,73 @@
...
+
+import textClassDataCollect from '@pipcook/pipcook-plugins-text-class-data-collect';
+import textClassDataAccess from '@pipcook/pipcook-plugins-text-csv-data-access';
+import bayesianClassiferModelLoader from '@pipcook/pipcook-plugins-bayesian-classifier-model-load';
+import bayesianClassifierModelTrain from '@pipcook/pipcook-plugins-bayesian-classifier-model-train';
+import classModelEvalute from '@pipcook/pipcook-plugins-bayesian-classifier-model-evaluate';
+import textClassLocalModelDeploy from '@pipcook/pipcook-plugins-text-class-local-model-deploy';
+import textClassEasDeploy from '@pipcook/pipcook-plugins-bayesian-classifier-model-eas-deploy'
+
+import { getEasParam, EasConfigI } from '../utils/utils';
```
yorkie:
```suggestion
import {getEasParam, EasConfigI} from '../utils/utils';
```

Consistent with the aboves.

packages/pipcook-app/src/models/textClassification.ts
```diff
@@ -0,0 +1,73 @@
...
+
+  constructor(model: string, metaData?: any, modelId?: string) {
+    this.model = model;
+    this.metaData = metaData;
+    if (modelId) {
+      this.modelId = modelId;
+    }
+  }
+
+  async _train(dataSource: string, trainInfo: any) {
```
yorkie: Why is this function async?

packages/pipcook-app/src/models/textClassification.ts
```diff
@@ -0,0 +1,73 @@
...
+  
+    const modelEvaluate = ModelEvaluate(classModelEvalute);
+
+    return {dataCollect, dataAccess, modelLoad, modelTrain, modelEvaluate};
+  }
+
+  async train(dataSource: string, trainInfo: any, 
+    predictServer=false, successCallback?: Function, errorCallback?: Function, saveModelCallback?: Function) {
+    
+    const {dataCollect, dataAccess, modelLoad, modelTrain, modelEvaluate} = await this._train(dataSource, trainInfo);
```
yorkie: Ditto.

packages/pipcook-app/src/models/textClassification.ts
```diff
@@ -0,0 +1,73 @@
...
+
+    const modelDeploy = ModelDeploy(textClassLocalModelDeploy)
+    
+    const runner = new PipcookRunner({
+      predictServer
+    });
+
+    runner.run([dataCollect, dataAccess, modelLoad, modelTrain, modelEvaluate, modelDeploy], successCallback, errorCallback, saveModelCallback)
```
yorkie: I prefer make the `_train` return this array to avoid unnecessary operations :)

packages/pipcook-cli/assets/example/pipcook-app-example/pipcook-textclass-app-test.js
```diff
@@ -0,0 +1,5 @@
+const pipcook = require('@pipcook/pipcook-app').default;
+
+const classifier = new pipcook.textClassification('bayes')
+
+classifier.train('http://ai-sample.oss-cn-hangzhou.aliyuncs.com/image_classification/datasets/textDataBinding.csv', {}, false);
```
yorkie: EOF

packages/pipcook-cli/assets/example/pipeline-example/pipeline-text-bayes-classification.js
```diff
@@ -1,10 +1,9 @@
 const {DataCollect, DataAccess, DataProcess, ModelLoad, ModelTrain, ModelEvaluate, ModelDeploy, PipcookRunner} = require('@pipcook/pipcook-core');
 
 const textClassDataAccess = require('@pipcook/pipcook-plugins-text-csv-data-access').default;
-const textClassDataProcess = require('@pipcook/pipcook-plugins-text-class-data-process').default
 const bayesianClassiferModelLoader = require('@pipcook/pipcook-plugins-bayesian-classifier-model-load').default;
 const bayesianClassifierModelTrain = require('@pipcook/pipcook-plugins-bayesian-classifier-model-train').default;
-const classModelEvalute = require('@pipcook/pipcook-plugins-model-evaluate').default;
+const classModelEvalute = require('@pipcook/pipcook-plugins-bayesian-classifier-model-evaluate').default;
```
yorkie: Ditto, let's write ES6 modules :)

packages/pipcook-plugins-bayesian-classifier-model-eas-deploy/src/assets/app.py
```diff
@@ -0,0 +1,80 @@
...
+    for t in range(0, len(all_words_list), 1):
+        if not all_words_list[t].isdigit() and all_words_list[t] not in stopwords_set and 1<len(all_words_list[t])<5:
+            feature_words.append(all_words_list[t])
+    return feature_words
+
+def processPredictData(data):
+    all_words_list = pickle.load(open('./model/feature_words.pkl', 'rb'))
+    word_cuts = []
+    for i in range(len(data)):
+      word_cuts.append(jieba.cut(data[i], cut_all=False) )
```
yorkie: Wrong indention

packages/pipcook-app/src/models/textClassification.ts
utkobe: there are two functions train and trainAndEasDeploy which share the function _train, the difference is that they are using different deploy plugin (one is local deploy, another one is eas deploy). So _train does not return the array with deploy.

utkobe:
@yorkie  Hey Yorkie, Thanks for the comment. I have fixed several style problems. Just one thing, all es6 module problems are related to example files. From my original opinion, I would like to enable user to just use `node <filename>` to run the examples. If I change it in es6 style, probably this would fail. For node version below 12, it requires --experimental-modules flag to run es6 module.

yorkie:
> From my original opinion, I would like to enable user to just use node <filename> to run the examples. If I change it in es6 style, probably this would fail. For node version below 12, it requires --experimental-modules flag to run es6 module.

I think we should do compatible works for commonJS if we want this.

utkobe:
> > From my original opinion, I would like to enable user to just use node  to run the examples. If I change it in es6 style, probably this would fail. For node version below 12, it requires --experimental-modules flag to run es6 module.
> 
> I think we should do compatible works for commonJS if we want this.

@yorkie I agree with that. I will do this to make it more nice. ^_^

PR #42, opened 2020-03-12
separate lint rules of pipcook-board-client and fix all warnings

utkobe:
This PR includes these changes related to lint
* Since the front-end web style can be quite different from node.js style, I just add separate lint rules to pipcook-board-client (just use standard [Alibaba Ice](https://ice.alibaba-inc.com/) rules)
* For now, close 'no-explicit-any' rule
* fix all other lint errors and warnings


.eslintignore
```diff
@@ -2,6 +2,10 @@ node_modules
...
 docs
 dist
+test
+pipcook_venv
 
 packages/pipcook-core/src/assets
-packages/pipcook-board-client/build
+packages/pipcook-cli/assets
+packages/pipcook-board-client
+packages/pipcook-board-server
```
yorkie: Missing EOF

package.json
```diff
@@ -17,7 +17,7 @@
   "scripts": {
     "postinstall": "lerna bootstrap",
     "build": "lerna run compile",
-    "lint": "eslint . --ext .ts --ext .js",
+    "lint": "eslint . --ext .ts --ext .js && lerna run lint",
```
yorkie: eslint should cover all packages.

packages/pipcook-board-client/.eslintcache
```diff
@@ -0,0 +1 @@
...,"/Users/queyue/Documents/work/pipcook/pipcook/packages/pipcook-board-client/src/pages/Dashboard/index.jsx",[],"/Users/queyue/Documents/work/pipcook/pipcook/packages/pipcook-board-client/src/pages/Dataset/index.jsx",[],"/Users/queyue/Documents/work/pipcook/pipcook/packages/pipcook-board-client/src/pages/Home/index.jsx",[],"/Users/queyue/Documents/work/pipcook/pipcook/packages/pipcook-board-client/src/pages/ImageClassification/index.jsx",[],"/Users/queyue/Documents/work/pipcook/pipcook/packages/pipcook-board-client/src/pages/Log/index.jsx",[],"/Users/queyue/Documents/work/pipcook/pipcook/packages/pipcook-board-client/src/pages/Model/index.jsx",[],"/Users/queyue/Documents/work/pipcook/pipcook/packages/pipcook-board-client/src/pages/NotFound/index.jsx",[],"/Users/queyue/Documents/work/pipcook/pipcook/packages/pipcook-board-client/src/pages/ObjectDetection/index.jsx",[],"/Users/queyue/Documents/work/pipcook/pipcook/packages/pipcook-board-client/src/pages/Status/index.jsx",[],"/Users/queyue/Documents/work/pipcook/pipcook/packages/pipcook-board-client/src/pages/TextClassification/index.jsx",[],"/Users/queyue/Documents/work/pipcook/pipcook/packages/pipcook-board-client/src/router.jsx",[]]
```
yorkie: Should remove the cache file?

packages/pipcook-board-client/.eslintignore
```diff
@@ -0,0 +1,13 @@
...
+demo/
+.ice/
+
+coverage/
+
+**/*-min.js
+**/*.min.js
+
+package-lock.json
+yarn.lock
```
yorkie: EOF

packages/pipcook-board-client/.eslintrc.js
```diff
@@ -0,0 +1,3 @@
+const { eslint } = require('@ice/spec');
+
+module.exports = eslint;
```
yorkie: EOF

yorkie:
```js
module.exports = require('@ice/spec').eslint;
```

packages/pipcook-board-client/package.json
```diff
@@ -33,7 +33,7 @@
     "start": "ice-scripts dev",
     "build": "ice-scripts build",
     "lint": "npm run eslint && npm run stylelint",
-    "eslint": "eslint --cache --ext .js,.jsx ./",
+    "eslint": "eslint --cache --ext .js,.jsx ./ --fix",
```
yorkie: `--fix` should not be included the npm command, run the following if someone wants to fix:

```sh
$ npm run eslint -- --fix
```

packages/pipcook-core/src/board/board.ts
```diff
@@ -27,7 +27,7 @@ export function getServer() {
  * @param fastify : fastify server instance
  */
 export function startServer(fastify: any) {
-  fastify.listen(7778, '0.0.0.0' ,function (err: Error, address: string) {
+  fastify.listen(7778, '0.0.0.0' ,function (err: Error) {
```
yorkie:
```suggestion
  fastify.listen(7778, '0.0.0.0', function (err: Error) {
```

package.json
yorkie: > Since the front-end web style can be quite different from node.js style, I just add separate lint rules to pipcook-board-client (just use standard Alibaba Ice rules)

So it's trying to add multiple lint rules, but this seems adding duplicated checks, how about  a scope way, like: `lerna run --scope @pipcook/pipcook-board-client lint`?

yorkie: And we should exclude the board-client source from https://github.com/alibaba/pipcook/blob/master/.eslintrc.js :)

packages/pipcook-board-client/.eslintcache
utkobe: done

package.json
utkobe: I exclude the board-client from .eslintignore. `lerna run lint` will not cause duplicate checks since only pipcook-board-client has lint command. Anyway, it makes sense to add scope to avoid any unexpected behaviour later.

packages/pipcook-board-client/.eslintrc.js
utkobe: This is the default file generated by [ice scaffold.](https://ice.alibaba-inc.com/) At this stage I think we can just remain this

yorkie:
Thanks for the hard work @utkobe :)

PR #46, opened 2020-03-20
meta: adjustment for builtin plugin directories

utkobe:
This PR includes:
* move plugins-related folder into subfolder 'plugins'
* remove prefix 'pipcook-plugins' for plugin folders
* fix tjs-node-gpu verison

@yorkie Please have a review please


packages/pipcook-app/package.json
```diff
@@ -10,7 +10,7 @@
   "scripts": {
     "test": "echo \"Error: no test specified\" && exit 1",
     "build": "npm run clean && npm run compile",
-    "clean": "rm -rf ./dist && rm -rf tsconfig.tsbuildinfo",
+    "clean": "rm -rf ./dist && rm -rf tsconfig.tsbuildinfo && rm -f package-lock.json",
```
yorkie: This makes a side effect, that we removed a git-managed file, why do we remove the package-lock.json?

yorkie:
And how about creating a shared `tsconfig.json` for all built-in plugins?

packages/pipcook-app/package.json
utkobe: Ah, this is my mistake. I modify this for test purpose. no good reasons to remove package-lock, I have modified this back

utkobe:
> And how about creating a shared `tsconfig.json` for all built-in plugins?

For lerna practise, I follow [some guides online](https://blog.logrocket.com/setting-up-a-monorepo-with-lerna-for-a-typescript-project-b6a81fe8e4f8/). I guess it's recommended to create separate tsconfig for each package and extends a common tsconfig since package should be regarded as independent pacakge? How do you think

yorkie:
Thanks for the guideline link, and how about tweaking our `tsconfig` inheritance as the following:

```
(tsconfig for independent plugin)
-------------------------------
tsconfig for common plugin
-------------------------------
tsconfig for root
```

And there are some notes for the above:

- independent plugin configuration is optional, but "extend" is required.
- common plugin config could be also used for our non-official plugin developer.

PR #48, opened 2020-03-20
core: export plugin constant.

Txiaozhe:


packages/pipcook-core/src/index.ts
```diff
@@ -9,4 +9,5 @@ export {PipcookRunner} from './core/core';
 export {createAnnotationFile, parseAnnotation, unZipData, downloadZip, getOsInfo, transformCsv,
   getDatasetDir, createAnnotationFromJson, getMetadata, getModelDir, convertPascol2CocoFileOutput, compressTarFile } from './utils/publicUtils';
 export {startBoard} from './board/board';
-export {PipcookComponentResult} from './types/component';
\ No newline at end of file
+export {PipcookComponentResult} from './types/component';
+export {PLUGINS} from './constants/plugins';
```
yorkie: We should export `constants` only at the top-level just like `constants.PLUGINS` :)

Txiaozhe: Yes I did, please take a look:)

packages/pipcook-core/src/index.ts
```diff
@@ -9,4 +9,14 @@ export {PipcookRunner} from './core/core';
...
+export {
+  DATACOLLECT,
+  DATAACCESS,
+  DATAPROCESS,
+  MODELLOAD,
+  MODELTRAIN,
+  MODELEVALUATE,
+  MODELDEPLOY,
+  ONLINETRAIN
+} from './constants/plugins';
```
yorkie: What I meant is:

```ts
import {PLUGINS} from './constants/plugins';
...
export constants = {
  PLUGINS,
};
```

Txiaozhe: awkward:(

yorkie: Never mind :p

PR #73, opened 2020-04-10
update and correct package.json

utkobe:
1. I found that currently the versions in package.json files are not in sync with npm
2. There are lots of not-used dependencies in package.json, just clean them up


packages/cli/bin/index.js
```diff
@@ -10,6 +10,11 @@ const start = require('../lib/start');
 const devPlugin = require('../lib/devPlugin');
 const dataset = require('../lib/dataset');
 
+const path = require('path');
+
+process.env.NODE_PATH = path.join(process.cwd(), 'node_modules');
+require("module").Module._initPaths();
```
yorkie, line 16: Why do we change the NODE_PATH?

utkobe: to change the order node searches node_modules so that node_modules in current working directory would be searched. (because pipcook-cli is installed globally but plugins are installed locally)

yorkie: The working directory `node_modules` should be in the search paths by default as:

```
NODE_MODULES_PATHS(START)
1. let PARTS = path split(START)
2. let I = count of PARTS - 1
3. let DIRS = [GLOBAL_FOLDERS]
4. while I >= 0,
   a. if PARTS[I] = "node_modules" CONTINUE
   b. DIR = path join(PARTS[0 .. I] + "node_modules")
   c. DIRS = DIRS + DIR
   d. let I = I - 1
5. return DIRS
```

utkobe:
Thanks for the review. CI is ok and I will just merge it

PR #74, opened 2020-04-10
boa: add support for linux

yorkie:
Let's using Boa at Linux 🚶


yorkie:
@utkobe All builds and tests are green, could you please have a look :)

yorkie:
Let's explain a bit more, this PR will increase the installation cost and make the following changes:

- It downloads the Python standalone via conda.
- It removes the Python from host requirements.
- don't support cross-compiling anymore, because we generate the "@rpath" at the host environment, but your still could change it via `patchelf` for your custom env.

packages/boa/tools/install-python.js
```diff
@@ -0,0 +1,37 @@
+'use strict';
+
+const { run, PLATFORM, ARCH } = require('./utils');
+const fs = require('fs');
+const path = require('path');
+
+const CONDA_DOWNLOAD_PREFIX = 'https://repo.anaconda.com/miniconda';
+let CONDA_DOWNLOAD_NAME = 'Miniconda3-latest';
```
Txiaozhe, line 8: By general convention, only constants are capitalized？

yorkie: IMO, the consistence with the CONDA_DOWNLOAD_PREFIX might be more important here :)

packages/boa/tools/install-python.js
```diff
@@ -0,0 +1,37 @@
...
+} else if (PLATFORM !== 'darwin') {
+  CONDA_DOWNLOAD_NAME += '-x86';
+}
+CONDA_DOWNLOAD_NAME = `${CONDA_DOWNLOAD_NAME}.sh`;
+
+if (!fs.existsSync(CONDA_DOWNLOAD_NAME)) {
+  run(`curl ${CONDA_DOWNLOAD_PREFIX}/${CONDA_DOWNLOAD_NAME} > ${CONDA_DOWNLOAD_NAME}`);
+}
+
+const CONDA_LOCAL_PATH = path.join(__dirname, '../.miniconda');
```
utkobe, line 29: I'm wondering if it's better to install .miniconda in process.cwd instead of under boa package. Otherwise if the user want to install some packages manually in conda env, it will be difficult to find this .miniconda directory. And in the meantime probably we can add a pipcook-cli command, for instance, pipcook install to let users install any packages later.

yorkie: Reasonable!

yorkie:
@utkobe @Txiaozhe Fixed the nits, and I introduced a new environment "BOA_CONDA_PREFIX" for:

- `@cwd`(**default**), install the conda into the working directory.
- `@package`, install the conda into the package directory, think about someone they don't care about Python.
- `String`, otherwise the absolute installation location is: `${BOA_CONDA_PREFIX}/.miniconda`.

utkobe:
Looks great.

PR #81, opened 2020-04-12
core: refine the filenames by following a consistent naming rule

yorkie:
Because we have no complete unit tests, so we have to make sure everything is ok in eyes :(


utkobe:
LGTM

PR #103, opened 2020-04-17
build: specify NAPI_VERSION to fix the node-addon-api issue

### LINKED ISSUE #99: "Pipcook init" node-gyp rebuild error
When i use node version 10 execute "pipcook init" will got error message:
image
I think maybe node version too old to got different API but pipcook not support?

yorkie:
This is to fix #99 which is a compiler error on Node.js 10.13.x~10.15.x, because there is a weird setup for NAPI_VERSION with NAPI_EXPERINMENT, and node-addon-api will not handle that, I have a root fix at https://github.com/nodejs/node-addon-api/pull/703.

But let's have this land to fix #99 ASAP.

### DIFF
--- a/packages/boa/binding.gyp
+++ b/packages/boa/binding.gyp
@@ -36,6 +36,7 @@
       "defines": [
         "NAPI_CPP_EXCEPTIONS",
         "NAPI_EXPERIMENTAL",
+        "NAPI_VERSION=6",
       ],
       "conditions": [
         ['OS=="mac"', {


legendecas:
See comment https://github.com/alibaba/pipcook/issues/99#issuecomment-615573899. I'm not convinced this is the root cause :(

yorkie:
Yea, this is not to fix the root cause of #99, this PR is to make Boa works on v10.13.x~v10.15.x, and using `NAPI_VERSION=6` is because Boa requires BigInt which is introduced at NAPI Version 6.

utkobe:
LGTM

Issue #105, opened 2020-04-17
plugin: help us to rewrite python-based files in boa


yorkie:
We still have some Python source code even though Boa is integrated, let's rewrite them:

- https://github.com/alibaba/pipcook/tree/master/packages/plugins/model-define/bayesian-model-define/src/assets
- https://github.com/alibaba/pipcook/tree/master/packages/plugins/model-evaluate/bayesian-model-evaluate/src/assets
- https://github.com/alibaba/pipcook/tree/master/packages/plugins/model-train/bayesian-model-train/src/assets

yorkie:
Thanks for your works @Txiaozhe, and there are the followings to be done:

- [ ] remove this directory https://github.com/alibaba/pipcook/tree/master/packages/plugins/model-define/bayesian-model-define/src/assets
- [ ] remove the python file: https://github.com/alibaba/pipcook/blob/master/packages/plugins/model-evaluate/bayesian-model-evaluate/src/assets/__init__.py
- [ ] https://github.com/alibaba/pipcook/tree/master/packages/plugins/model-train/bayesian-model-train/src/assets

yorkie:
I think this can be closed now, thanks for your works @Txiaozhe.

PR #107, opened 2020-04-19
boa: try rewrite python based file.

Txiaozhe:
This directory for this script.js file is temporary, and this file should be moved to right path.

Have some problems:
1. how to express the syntax for python use js, such as:
```python
for  (word in words) {
  ...
}
```
In my way I will use 'for of'/for/etc.

2. how to express the data structure, such as set:

use builtins:
```js
const {set} = boa.builtins() 
const s = set()
```
or use js native:
```js
const s = new Set()
```

3. some methods looks like builtin should be implemented in unison， such string.strip()

4. Can I  use some async function to express python based function?


Txiaozhe:
From the docs for boa, may use 'boa.eval'

yorkie:
As I have mentioned at #105, you should move this files in corresponding plugins.

yorkie:
@Txiaozhe There are some warnings still, could you please fix them?

Txiaozhe:
```python
#coding: utf-8
import os
import time
import random
import jieba
import sklearn
from sklearn.naive_bayes import MultinomialNB
import numpy as np
import csv
import pickle
```

```js
const boa = require('@pipcook/boa');

const os = boa.import('os');
const time = boa.import('timer');
const random = boa.import('random');
const jieba = boa.import('jieba');
const sklearn = boa.import('sklearn');
const { MultinomialNB } = boa.import('sklearn.naive_bayes');
const csv = boa.import('csv');
const pickle = boa.import('pickle');
```
warning: 'os' is assigned a value but never used  ; etc.
'os' imported in python file, so I import it in js, so~

yorkie:
They should be removed now :)

yorkie:
Now let's use your file at https://github.com/alibaba/pipcook/blob/master/packages/plugins/model-define/bayesian-model-define/src/index.ts, and the Python source could be removed, too.

packages/plugins/model-define/bayesian-model-define/src/script.js
```diff
@@ -0,0 +1,89 @@
+'use strict';
+
+const boa = require('@pipcook/boa');
+
+const jieba = boa.import('jieba');
+const { MultinomialNB } = boa.import('sklearn.naive_bayes');
+const pickle = boa.import('pickle');
+
+const { open } = boa.builtins();
```
yorkie, line 9: There is no necessary to use open instead of Node.js's file system APIs.

packages/plugins/model-define/bayesian-model-define/src/script.js
```diff
@@ -0,0 +1,89 @@
...
+
+const fs = require('fs');
+const readline = require('readline');
+
+function strip (str) {
```
yorkie, line 14:
```suggestion
function strip(str) {
```

packages/plugins/model-define/bayesian-model-define/src/script.js
```diff
@@ -0,0 +1,89 @@
...
+  });
+};
+
+exports.getBayesModel = function () {
+  const classifier = MultinomialNB();
+  return classifier;
+};
+
+exports.loadModel = function (filepath) {
+  const classifier = pickle.load(open(filepath, 'rb'));
```
yorkie, line 87:
```suggestion
  return pickle.load(open(filepath, 'rb'));
```

packages/plugins/model-define/bayesian-model-define/src/script.js
```diff
@@ -0,0 +1,89 @@
...
+  return MakeWordsSet(stopwords_file).then((stopwords_set) => {
+    const feature_words = words_dict(all_words_list, stopwords_set);
+    return word_cut.map((text) => {
+      return text_features(text, feature_words);
+    });
+  });
+};
+
+exports.getBayesModel = function () {
+  const classifier = MultinomialNB();
```
yorkie, line 82: Ditto.

Txiaozhe:
fixed.

yorkie:
Looks good to me, may I ask you to run the following to valid the pipeline is not broken:

```sh
$ sh run_pipeline.sh mnist-image-classification
```

yorkie:
And @utkobe could you please chime this in?

packages/plugins/model-define/bayesian-model-define/src/script.ts
```diff
@@ -0,0 +1,88 @@
...
+
+  return MakeWordsSet(stopwords_file).then((stopwords_set) => {
+    const feature_words = words_dict(all_words_list, stopwords_set);
+    return word_cut.map((text: string) => {
+      return text_features(text, feature_words);
+    });
+  });
+};
+
+export const getBayesModel = function (): any {
```
yorkie, line 82: Could you please remove the any type?

packages/plugins/model-define/bayesian-model-define/src/script.ts
```diff
@@ -0,0 +1,88 @@
...
+  return MultinomialNB();
+};
+
+export const loadModel = function (filepath: string): any {
```
yorkie, line 86: Ditto.

Txiaozhe:
> Looks good to me, may I ask you to run the following to valid the pipeline is not broken:
> 
> ```shell
> $ sh run_pipeline.sh mnist-image-classification
> ```

Okay, pipline is running.

Txiaozhe:
And some doubt:
This two functions:
```js
function strip(str: string): string {
  return str.replace(/(^\s*)|(\s*$)/g, '');
}

function isdigit(str: string): boolean {
  return /^\d+$/.test(str);
}
```
They are looks in common use, could they write into a lib?

yorkie:
Sorry, could you please run this one: https://github.com/alibaba/pipcook/blob/master/example/pipelines/text-bayes-classification.json, mnist is not related with this PR.

Txiaozhe:
> Sorry, could you please run this one: https://github.com/alibaba/pipcook/blob/master/example/pipelines/text-bayes-classification.json, mnist is not related with this PR.

OO

Txiaozhe:
Pipline was completed.

yorkie:
Good job, and it's good to have this in the list: https://github.com/alibaba/pipcook/blob/master/.github/workflows/pipeline.yml#L14.

packages/boa/requirements.txt
```diff
@@ -1,2 +1,3 @@
 numpy==1.17.4
-
+jieba==0.42.1
+sklearn==0.0
```
yorkie, line 3: I'm not sure if we should do this here :(

Txiaozhe: If not, it will throw error: 'not fount module 'jieba'' although I have 'jieba' use pip in my local.
And it means 'jieba' must install into dir: '.miniconda'?

utkobe: Yes, i think we can just keep jieba and sklearn here in requirements. After we have completed plugin runtime, this can be removed in the future.

utkobe:
The whole staff looks good to me. Thanks for the work. @Txiaozhe @yorkie

packages/boa/requirements.txt
Txiaozhe: Thanks and LGTM :)

yorkie: Resolved.

yorkie:
Merged, thank you @Txiaozhe.

PR #109, opened 2020-04-20
correct the generation of templates for plugins for development

utkobe:
As @shadowkun reflects, currently the generation of templates for plugins in developer mode are in errors, just to fix them and change the interface to the up-to-date interfaces in pipcook-core


packages/cli/assets/pluginPackage/src/dataAccess.ts
```diff
@@ -1,8 +1,8 @@
-import { UniformSampleData, OriginSampleData, ArgsType, parseAnnotation, DataAccessType} from '@pipcook/pipcook-core';
+import { UniDataset, ArgsType, DataAccessType} from '@pipcook/pipcook-core';
```
Txiaozhe, line 1: coincident space between '{' and '}'

packages/cli/assets/pluginPackage/src/dataCollect.ts
```diff
@@ -1,7 +1,7 @@
-import {OriginSampleData, ArgsType, DataCollectType} from '@pipcook/pipcook-core';
+import {ArgsType, DataCollectType} from '@pipcook/pipcook-core';
```
Txiaozhe, line 1: ditto

yorkie:
There is a conflicting file, @utkobe could you please have a fix?

PR #114, opened 2020-04-21
doc: siderbar should be in absolute

yorkie:


JihangGuo:
Changed to absolute path, no problem

PR #117, opened 2020-04-22
move deploy/predict feature into pipcook-core

utkobe:
This PR includes
- remove deploy plugins for now and pack the deploy/predict logic in core
- according change the logic in data-process plugin so that the data process plugin will be used to handle single sample. (This is both benefit for training step and prediction step to pre-process data)
- fix some problems in bayeisan model define (use python open instead nodejs's)


packages/core/src/assets/predict.js
```diff
@@ -0,0 +1,41 @@
+const modelDefine = require('./modelDefine').default;
+const fastify = require('fastify')({ logger: true });
```
yorkie, line 2: Could we remove the fastify dependence and only export as a library? This makes integration to other frameworks to be easy more :)

packages/plugins/model-define/bayesian-model-define/src/script.ts
```diff
@@ -71,18 +72,16 @@ export const processPredictData = function (data: any, all_words_list_path: stri
...
-  return MakeWordsSet(stopwords_file).then((stopwords_set) => {
-    const feature_words = words_dict(all_words_list, stopwords_set);
-    return word_cut.map((text: string) => {
-      return text_features(text, feature_words);
-    });
-  });
+  const stopwords_set = await MakeWordsSet(stopwords_file);
+  const feature_words = words_dict(all_words_list, stopwords_set);
+  const feature_list = [ text_features(word_cut, feature_words) ];
+  return feature_list;
```
Txiaozhe, line 78: Could it return [ text_features(word_cut, feature_words) ]; directly？

utkobe: done

packages/core/src/assets/predict.js
utkobe: Yes, i'm thinking we could remove fastify dependency here, and just leave the pure prediction function so that it's easy to intergrate to other systems. But in the meantime, we can add a command in cli such as `pipcook try(predict) ...` to use fastify quickly run a server locally (just for beginners to quickly check results intuitively). Does this make sense

yorkie: How about `pipcook serve`? Because the `try` sub command might be reversed for MLApp.

utkobe: done~

packages/core/src/assets/template-package.json
```diff
@@ -0,0 +1,15 @@
+{
+  "name": "pipcook-deploy-template",
+  "version": "0.0.1",
+  "description": "",
+  "main": "main.js",
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1"
+  },
+  "author": "pipcook",
+  "license": "ISC",
```
yorkie, line 10: Change this to our license :)

packages/core/src/assets/template-package.json
```diff
@@ -0,0 +1,15 @@
+{
+  "name": "pipcook-deploy-template",
+  "version": "0.0.1",
+  "description": "",
+  "main": "main.js",
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1"
+  },
+  "author": "pipcook",
```
yorkie, line 9: Ditto

yorkie:
And we need to update our documentations to remove model-deploy related stuffs in this PR.

packages/cli/lib/serve.js
```diff
@@ -0,0 +1,39 @@
...
+  fastify.post('/', async (req) => {
+    const result = await predictFunc(req.body.data);
+    return {
+      result: result
+    };
+  });
+
+  const start = async () => {
+    try {
+      await fastify.listen(7687);
```
yorkie, line 30: How about an `-p --port` argument for this?

utkobe:
> And we need to update our documentations to remove model-deploy related stuffs in this PR.

Have done that

PR #128, opened 2020-04-24
Release workflow

utkobe:
This is to add a workflow for releasing the packages. The workflow will be triggered when a tag with v* is created. 

In the future, when a new pipcook version is release, we will use github release to create a new one with a new tag and trigger this workflow


.github/workflows/release.yml
```diff
@@ -0,0 +1,34 @@
...
+      - name: publish npm Packages
+        run: |
+          echo "//registry.npmjs.org/:_authToken=${{ secrets.npm_token }}" > ~/.npmrc
+          git config --global user.name 'pipcook'
+          git config --global user.email 'queyue.crk@alibaba-inc.com'
+          npm run release
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          NODE_AUTH_TOKEN: ${{secrets.npm_token}}
+          NPM_AUTH_TOKEN: ${{ secrets.npm_token }}
```
yorkie, line 34: Missing EOF.

packages/boa/.npmignore
```diff
@@ -1,2 +1,12 @@
 .gitignore
 .CONDA_INSTALL_DIR
+.dataset
+.data
+.checkpoints
+
+**/__pycache__
+.idea/.DS_Store/**
+.CONDA_INSTALL_DIR
```
yorkie, line 9: This seems duplicated with the line 2.

utkobe:
It turns out green and I will merge this one

PR #133, opened 2020-04-25
plugin, core: fail the pipeline when evaluation is not pass

### LINKED ISSUE #129: Checking model accuracy for every merge.

yorkie:
This fixes #129 by introducing a pass field on `EvaluateResult` interface, the core will throw a corresponding `EvaluateError` when this field is not set true.

### DIFF
--- a/packages/plugins/model-evaluate/object-detection-detectron-model-evaluate/src/index.ts
+++ b/packages/plugins/model-evaluate/object-detection-detectron-model-evaluate/src/index.ts
@@ -14,17 +14,12 @@ const detectronModelEvaluate: ModelEvaluateType = async (data: CocoDataset, mode
     const { COCOEvaluator, inference_on_dataset } = boa.import('detectron2.evaluation');
     const { build_detection_test_loader } = boa.import('detectron2.data');
 
-    register_coco_instances("test_dataset", {}, data.testAnnotationPath, path.join(data.testAnnotationPath, '..'));
-    cfg.DATASETS.TEST = [ "test_dataset" ];
+    register_coco_instances('test_dataset', {}, data.testAnnotationPath, path.join(data.testAnnotationPath, '..'));
+    cfg.DATASETS.TEST = [ 'test_dataset' ];
 
-    const evaluator = COCOEvaluator("test_dataset", cfg, false, boa.kwargs({ output_dir: modelDir }));
-    const val_loader = build_detection_test_loader(cfg, "val_dataset");
-    const result = inference_on_dataset(trainer.model, val_loader, evaluator);
-
-    return {
-      pass: result > expectResult,
-      result
-    };
+    const evaluator = COCOEvaluator('test_dataset', cfg, false, boa.kwargs({ output_dir: modelDir }));
+    const val_loader = build_detection_test_loader(cfg, 'val_dataset');
+    return inference_on_dataset(trainer.model, val_loader, evaluator);
   }
 };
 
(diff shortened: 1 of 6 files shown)


packages/plugins/model-evaluate/object-detection-detectron-model-evaluate/src/index.ts
```diff
@@ -22,13 +18,12 @@ const detectronModelEvaluate: ModelEvaluateType = async (data: CocoDataset, mode
...
     const evaluator = COCOEvaluator("test_dataset", cfg, false, boa.kwargs({ output_dir: modelDir }));
-      
     const val_loader = build_detection_test_loader(cfg, "val_dataset");
-
-    const val_result = inference_on_dataset(trainer.model, val_loader, evaluator);
+    const result = inference_on_dataset(trainer.model, val_loader, evaluator);
 
     return {
-      result: val_result
+      pass: result > expectResult,
```
Txiaozhe, line 25: may use '>=' ?

Txiaozhe:
> This fixes #129 by introducing a pass field on `EvaluateResult` interface, the core will throw a corresponding `EvaluateError` when this field is not set true.

It is a good idea to check expect value for every training~

packages/plugins/model-evaluate/object-detection-detectron-model-evaluate/src/index.ts
```diff
@@ -22,13 +18,12 @@ const detectronModelEvaluate: ModelEvaluateType = async (data: CocoDataset, mode
     cfg.DATASETS.TEST = [ "test_dataset" ];
 
     const evaluator = COCOEvaluator("test_dataset", cfg, false, boa.kwargs({ output_dir: modelDir }));
-      
     const val_loader = build_detection_test_loader(cfg, "val_dataset");
-
-    const val_result = inference_on_dataset(trainer.model, val_loader, evaluator);
+    const result = inference_on_dataset(trainer.model, val_loader, evaluator);
```
utkobe, line 22: inference_on_dataset method should return a dict which contains AP50, AP75 for each class and MAP (depends on config). So I think it's not valid when use `result > expectResult`

utkobe:
Just wondering if it's a good idea to make 'pass' attribute necessary. Probably the core will throw errors only if the pass attribute is set to false. Otherwise when it's true or undefined, the core will think it's ok?

packages/plugins/model-evaluate/object-detection-detectron-model-evaluate/src/index.ts
yorkie: Thanks @utkobe, But I can't find any about the keys list at https://detectron2.readthedocs.io/_modules/detectron2/evaluation/evaluator.html#DatasetEvaluator.evaluate, is there the kind of list?

yorkie:
> Just wondering if it's a good idea to make 'pass' attribute necessary. Probably the core will throw errors only if the pass attribute is set to false. Otherwise when it's true or undefined, the core will think it's ok?

This depends on how do we look at our evaluation plugins, this PR's viewpoint is that the evaluate outputs a `pass` value, and treat the map data as the side-effect. And your proposal is going to output the map data and with the `pass` value as side-effect.

BTW, IMOH this latter way is not that pure enough, because in fact the core will use `pass` value only, however we make a useless map object be required, is that weird?

packages/plugins/model-evaluate/object-detection-detectron-model-evaluate/src/index.ts
utkobe: could have a look at https://detectron2.readthedocs.io/_modules/detectron2/evaluation/coco_evaluation.html#COCOEvaluator.__init__    essentially it will use pycocotools to evaluate standard metrics required by coco

yorkie:
@utkobe I have fixed all the nits, and tweaked the negative check for `pass` :)

PR #134, opened 2020-04-25
cli: checking config file exist or not.

Txiaozhe:
Could I add unit test for cli in /packages/cli/test, and use 'mocha' as tool.

### DIFF
--- a/packages/cli/lib/start.js
+++ b/packages/cli/lib/start.js
@@ -11,7 +11,7 @@ const start = async (filename) => {
   }
 
   if (!fs.existsSync(filename)) {
-    spinner.fail(`Config file: '${filename}' not exist.`);
+    spinner.fail(`${filename} not exists`);
     return;
   }
 
--- a/packages/cli/lib/start.js
+++ b/packages/cli/lib/start.js
@@ -1,11 +1,17 @@
 const fork = require('child_process').fork;
 const ora = require('ora');
 const spinner = ora();
+const fs = require('fs')
 const path = require('path');
 
 const start = async (filename) => {
   if (!filename) {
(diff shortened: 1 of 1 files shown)


yorkie:
> Could I add unit test for cli in /packages/cli/test, and use 'mocha' as tool.

Sorry, we use jasmine as the testing framework.

packages/cli/lib/start.js
```diff
@@ -9,6 +10,11 @@ const start = async (filename) => {
     return;
   }
 
+  if (!fs.existsSync(filename)) {
+    spinner.fail(`Config file: '${filename}' not exist.`);
```
yorkie, line 14:
```suggestion
    spinner.fail(`${filename} not exists`);
```

Txiaozhe:
> > Could I add unit test for cli in /packages/cli/test, and use 'mocha' as tool.
> 
> Sorry, we use jasmine as the testing framework.

So could I add unit test for cli in /packages/cli/test?

yorkie:
> So could I add unit test for cli in /packages/cli/test?

I'm not sure if this is the right time, because cli is still not using the TypeScript, maybe the cli unit tests can be added after that.

Txiaozhe:
> > So could I add unit test for cli in /packages/cli/test?
> 
> I'm not sure if this is the right time, because cli is still not using the TypeScript, maybe the cli unit tests can be added after that.

OO, it is reasonable :)

yorkie:
And also that would be great if you help us to rewrite cli in TypeScript :)

Txiaozhe:
> And also that would be great if you help us to rewrite cli in TypeScript :)

I’d love to and could you please create an issue and assign it to me? :)

PR #139, opened 2020-04-27
core: correct the coco's id in public utils

utkobe:
the converter from pascal to coco has some problems where the id of image is not corresponding to image_id in annotations. Just to fix it


yorkie:
Could we add a simple test to make sure this will happen again?

Txiaozhe:
May use mocha as unit test framework?

utkobe:
> May use mocha as unit test framework?

Have already used jasmine before if you have a check of npm run test command under root

Issue #72, opened 2020-04-09
core: remove `any` types as possible


yorkie:
The `any` is not recommended in TypeScript, feel free to open PR to help us remove one or them :)

yorkie:
Still not get fixed fully :)

FeelyChau:
Almost all `any` types have been fixed.

PR #140, opened 2020-04-27
core: remove `any` types for `lifecycle.ts`

liangzr:
fix [#72](https://github.com/alibaba/pipcook/issues/72)

**details**
- Replace `any` for pipcook plugin params with the existing `PipObject`
- Remove `any` that could be inferred by other definition
- Use TypeScript built-in generic types first

### DIFF
--- a/packages/core/src/types/other.ts
+++ b/packages/core/src/types/other.ts
@@ -10,9 +10,7 @@ export interface DeploymentResult {
   extraData: any;
 } 
 
-export interface PipObject {
-  [key: string]: any;
-}
+export type PipObject = Record<string, any>
 
 export interface EvaluateResult {
   [key: string]: any;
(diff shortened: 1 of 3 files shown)


packages/core/src/types/other.ts
```diff
@@ -10,9 +10,7 @@ export interface DeploymentResult {
   extraData: any;
 } 
 
-export interface PipObject {
-  [key: string]: any;
-}
+export type PipObject = Record<string, any>
```
yorkie, line 13: Good improvements, just figure out another "good first issue": use TypeScript `Record` instead of custom object.

/cc @utkobe

yorkie: Added: https://github.com/alibaba/pipcook/issues/142

PR #148, opened 2020-04-28
boa: support toJSON on proxy object

### LINKED ISSUE #146: boa: the boa object cannot be stringfied
Boa python wrapper object cannot be converted to string using JSON.stringify. It's good to support this transformation or provide a method like .toJson() to easily convert to JS native object.

For example

```js
const boa = require('../packages/boa');
const obj = np.array([1,2,3]);
console.log(JSON.stringify(obj));
```

The above codes will throw error

```
TypeError: Converting circular structure to JSON
    --> starting at object with constructor 'Object'
    |     property '__handle__' -> object with constructor 'boa.Object'
    --- property '__wrapped__' closes the circle
    at JSON.stringify (<anonymous>)
```

yorkie:
This fixes #146.

### DIFF
--- a/packages/boa/lib/index.js
+++ b/packages/boa/lib/index.js
@@ -47,6 +47,32 @@ function copy(T) {
   return fn.invoke(asHandleObject(T));
 }
 
+function list(T) {
+  return builtins.__getitem__('list').invoke(asHandleObject(T));
+}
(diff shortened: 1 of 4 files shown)


utkobe:
LGTM

PR #150, opened 2020-04-28
core: add costa and integrate to daemon

yorkie:
This PR tries to export a class `PluginRT` which provides the following methods:

- `fetch()`, fetch the metadata for given plugin.
- `install()`, install plugin by the fetched information.
- `run()`, run the given plugin with arguments.
- `test()`, it's used for helping to test plugins in an easy way.

And internally, 

- it initializes a NPM packages to manage plugins.
- it uses Node.js's child process for managing plugin clusters. (Maybe slow in bootstrap)


Txiaozhe:
Does need uninstall()?

yorkie:
Sure.

packages/boa/package.json
```diff
@@ -4,6 +4,7 @@
   "description": "Use Python modules seamlessly in Node.js",
   "main": "lib/index.js",
   "scripts": {
+    "clean": "rm -rf ./build && rm -rf ./node_modules",
```
Txiaozhe, line 7: Does need delete node_modules dir?

packages/costa/src/index_test.ts
```diff
@@ -0,0 +1,21 @@
...
+  const collectCsv = await rt.fetch('../plugins/data-collect/csv-data-collect');
+  await rt.install(collectCsv);
+  await rt.run(collectCsv, {
+    url: 'http://ai-sample.oss-cn-hangzhou.aliyuncs.com/image_classification/datasets/textClassification.zip'
+  });
+  // await rt.run('@pipcook/plugins-csv-data-access', {
+  //   labelColumn: 'output'
+  // });
+  // await rt.run('@pipcook/plugins-bayesian-model-define');
+  // await rt.run('@pipcook/plugins-bayesian-model-train');
```
Txiaozhe, line 20: Does need delete useless code?

yorkie: This PR is in progress, please skip it for now :)

Txiaozhe: OO

packages/costa/src/proto.ts
```diff
@@ -0,0 +1,23 @@
...
+
+export enum PluginOperator {
+  START = 0x30,
+  WRITE = 0x31,
+  READ = 0x32,
+  COMPILE = 0x100
+}
+
+export class PluginProto {
+  op: PluginOperator;
```
Txiaozhe, line 14: Maybe 'operator' is better just like 'message' below?

packages/costa/src/proto.ts
```diff
@@ -0,0 +1,23 @@
...
+  message: PluginMessage;
+
+  static stringify(op: PluginOperator, message: PluginMessage): string {
+    return JSON.stringify({ op, message });
```
Txiaozhe, line 18: Ditto?

packages/costa/src/client.ts
```diff
@@ -0,0 +1,42 @@
...
+const handlers: MessageHandler = {
+  [PluginOperator.START]: (proto: PluginProto) => {
+    pong();
+  },
+  [PluginOperator.WRITE]: async (proto: PluginProto) => {
+    const { event, params } = proto.message;
+    debug(`receive an event ${event}`);
+    if (event === 'load') {
+      debug(`start loading plugin ${params[0]}`);
+      const plugin = require(params[0]).default;
```
Txiaozhe, line 23: May need error catching for require?

packages/costa/src/proto.ts
yorkie: The message is called PluginMessage, either.

packages/costa/src/proto.ts
yorkie: I dunno what do you mean, maybe in details?

packages/boa/requirements.txt
```diff
@@ -1,3 +1,3 @@
 numpy==1.17.4
 jieba==0.42.1
-sklearn==0.0
\ No newline at end of file
+sklearn==0.0
```
utkobe, line 3: it is recommended to use scikit-learn and specify specific version instead :)

packages/cli/src/actions/init.ts
```diff
@@ -0,0 +1,85 @@
...
+  // Install the daemon and pipboard.
+  // use ~/.pipcook as PIPCOOK_DIR
+  const PIPCOOK_DIR = path.join(os.homedir(), '.pipcook');
+  const DAEMON_DIR = path.join(PIPCOOK_DIR, 'server');
+  const BOARD_DIR = path.join(PIPCOOK_DIR, 'pipboard');
+
+  try {
+    await fse.ensureDir(DAEMON_DIR);
+    await fse.ensureDir(BOARD_DIR);
+    npmInstall(npmClient, daemonPackage, beta, DAEMON_DIR, npmInstallEnvs);
```
utkobe, line 75: Remember we cannot use tnpm to install daemon because the midway version is not correct with tnpm. Therefore I made a check here that if user choose tnpm, i will replace it with npm

packages/daemon/src/service/pipeline.ts
```diff
@@ -22,6 +24,14 @@ function getIdOrName(id: string): QueryParams {
   return validate(id) as boolean ? { id } : { name: id };
 }
 
+function execAsync(cmd: string, opts: ExecOptions): Promise<string> {
+  return new Promise((resolve, reject): void => {
+    exec(cmd, opts, (err: ExecException, stdout: string, stderr: string) => {
+      err == null ? resolve(stdout) : reject(err);
```
Txiaozhe, line 29: how about: err ? resolve(...) : reject(...)

yorkie: This is an incorrect expression, we should not resolve when there is no errors.

Txiaozhe: Awkward:(, I mean !err ? resolve(...) : reject(...)

yorkie: I don't think it is necessary, you can explain what is wrong with the current way.

Txiaozhe: Nothing else~

yorkie: Resolved, thank you :)

packages/boa/requirements.txt
yorkie: Okay, changed to `scikit-learn@0.23.1`.

yorkie:
I marked this as ready for review, and this implements the following functions finally:

- add a new module "@pipcook/costa" as the plugin plugin, which provide the runtime and managing packages.
- support plugin management in daemon.
- support plugin management in cli.

packages/plugins/model-train/bayesian-model-train/src/script.ts
```diff
@@ -119,17 +118,17 @@ export const processPredictData = function (data: any, all_words_list_path: stri
...
-export const getBayesModel = function() {
-  return MultinomialNB();
+  return boa.with(open(filepath, 'wb'), (f: any) => {
+    return pickle.dump(feature_words, f);
+  });
 };
 
-export const saveBayesModel = function(classifier: any, filepath: string) {
-  pickle.dump(classifier, open(filepath, 'wb'));
+export const saveBayesModel = async function(classifier: any, filepath: string) {
```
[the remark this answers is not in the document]
yorkie: `boa.with` returns a Promise object, so it might be reasonable as async, or specify the returned value as Promise.

packages/cli/src/actions/init.ts
yorkie: Do we need to avoid tnpm?

utkobe: Not quite sure about this. As far as i know, lots of developers inside Alibaba like using tnpm which is much faster with company intranet.

yorkie: Ok, we should remove the tnpm on our community edition :)

utkobe:
It seems that only package.json for text-classifaction has 'pipcook' field. And costa will check this filed existence.

yorkie:
> It seems that only package.json for text-classifaction has 'pipcook' field. And costa will check this filed existence.

Yes, this is just for reviewing only, I will update other plugins when we are ready to land :)

packages/daemon/src/service/pipeline.ts
```diff
@@ -1,170 +1,259 @@
...
+    const modelEvaluate = await costa.fetchAndInstall(pipeline.modelEvaluate, cwd);
+    const output = await runnable.start(modelEvaluate, dataset, model, getParams(pipeline.modelEvaluateParams, {
+      modelDir: modelPath
+    }));
+
+    const result = await runnable.valueOf(output) as EvaluateResult;
+    job.evaluateMap = JSON.stringify(result);
+    job.evaluatePass = result.pass;
+    job.endTime = Date.now();
+    job.status = 1;
```
utkobe, line 208: We need a error handling so that if errors occur, job.status will be 2

yorkie:
Supported the server-sent events via this library: https://github.com/EventSource/eventsource. And our implementations are:

- client(cli): https://github.com/alibaba/pipcook/blob/create/plugin-runtime/packages/cli/src/actions/start.ts#L37-L56
- server(daemon): https://github.com/alibaba/pipcook/blob/create/plugin-runtime/packages/daemon/src/app/controller/job.ts#L48-L68

And note that SSE on server-side requires the method to be GET only.

yorkie:
SSE is really such fit with our needs, it's lightweight and standardized, which is supported both on Web and Node.js, so I think we will moved our updates to SSE in this PR later:

1. daemon logs includes.
2. daemon state updates from a request.

packages/plugins/model-define/detectron-fasterrcnn-model-define/package.json
```diff
@@ -26,5 +26,17 @@
...
+  "pipcook": {
+    "category": "modelDefine",
+    "datatype": "image",
+    "method": "detect"
+  },
+  "conda": {
+    "python": "3.7",
+    "dependencies": {
+      "torch": "*",
+      "torchvision": "*"
```
utkobe, line 39: does this support pip install from git. For example, to install detectron, we need to run `pip install 'git+https://github.com/facebookresearch/detectron2.git'`

yorkie: Will support :)

yorkie: Supported at https://github.com/alibaba/pipcook/commit/775b1c496f0b658f50152f32f98fce3fcd12bd04#diff-32df0838fb24296b5d38739a7cf1b1fdR40 :)

yorkie:
All pipelines could be ran successfully :)

packages/costa/src/runtime.ts
```diff
@@ -0,0 +1,265 @@
...
+   * @param name the plugin package name.
+   * @param force install from new anyway.
+   * @param pyIndex the index mirror to install python packages.
+   */
+  async install(pkg: PluginPackage, force = false, pyIndex?: string): Promise<boolean> {
+    // check if the pkg is installed
+    if ((await this.isInstalled(pkg.name)) && !force) {
+      debug(`skip install "${pkg.name}" because it already exists`);
+      return true;
+    }
```
liangzr, line 108: Does this operation conflict with npm's strategy? It also ignores the package version specified by the plugin, maybe this is an unnecessary step?

packages/costa/src/runtime.ts
```diff
@@ -0,0 +1,265 @@
...
+   * Uninstall the given plugin by name.
+   * @param name the plugin package name.
+   */
+  async uninstall(name: string): Promise<boolean> {
+    if (!await this.isInstalled(name)) {
+      debug(`skip uninstall "${name}" because it not exists.`);
+      return false;
+    }
+    await remove(path.join(this.options.installDir, 'node_modules', name));
+    return true;
```
liangzr, line 178: Using `npm uninstall <package-name>` will be better, remove won't purge its dependencies at the same time.

packages/costa/src/index.ts
```diff
@@ -0,0 +1,194 @@
...
+  _rev: string;
+  name: string;
+  'dist-tags': {
+    beta: string;
+    latest: string;
+  };
+  versions: Record<string, NpmPackage>;
+}
+
+export declare class PluginRunnable {
```
liangzr, line 132: Same problem as `CostaRuntime` :P

packages/costa/src/proto.ts
```diff
@@ -0,0 +1,39 @@
...
+  READ = 0x32
+}
+
+/**
+ * Plugin Exchange Protocol.
+ */
+export class PluginProto {
+  op: PluginOperator;
+  message: PluginMessage;
+
```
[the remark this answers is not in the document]
yorkie: https://github.com/alibaba/pipcook/blob/create/plugin-runtime/packages/costa/src/client.ts#L20, it's used here.

packages/costa/src/runtime.ts
```diff
@@ -0,0 +1,265 @@
...
+  });
+}
+
+function createRequirements(name: string, config: CondaConfig): string[] {
+  const deps = [];
+  for (let k in config.dependencies) {
+    const v = config.dependencies[k];
+    if (v === '*') {
+      deps.push(k);
+    } else if (v.startsWith('git+https://') === true) {
```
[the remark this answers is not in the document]
yorkie: Yea, I just keep consistent with the above `v === '*'`.

packages/costa/src/index.ts
```diff
@@ -0,0 +1,194 @@
...
+}
+
+/**
+ * The Costa runtime is for scheduling plugins and management.
+ */
+export declare class CostaRuntime {
+  /**
+   * The runtime config.
+   */
+  public options: RuntimeOptions;
```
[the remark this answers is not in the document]
yorkie: Nope, I still leave this declaration here is to be API reference :)

packages/costa/src/runtime.ts
yorkie: Actually I used `npm uninstall`, but it will cleanup `node_modules` under the plugin directory.

packages/costa/src/runtime.ts
yorkie: NPM strategy is not transparent, and this step is required at Pipcook, so we need to make sure it works specifically.

packages/costa/src/index.ts
yorkie: See https://github.com/alibaba/pipcook/pull/150#discussion_r432232947.

packages/costa/src/proto.ts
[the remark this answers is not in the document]
liangzr: The above place only uses input parameters that have the same naming, not class members. Actually, static methods can't be accessing variable from instance

yorkie:
> Except for the following comments, another common problem is many `@types/*` packages are listed in `dependencies`, this will slow down the installation on production environment

Should we move `@types` to `devDependencies`?

liangzr:
> Should we move `@types` to `devDependencies`?

Maybe we could leave this in another PR, I’ll do some tests :)

utkobe:
LGTM :)

liangzr:
LGTM~ :)

Txiaozhe:
Celebrations~

PR #155, opened 2020-04-30
core: remove `any` of plugin's lifecycle

liangzr:
fix [#72](https://github.com/alibaba/pipcook/issues/72)

**details**
- Add custom generic tool `PromisedValueOf`
- Remove `any` of observable in plugin's lifecycle

### DIFF
--- a/packages/core/src/types/utility.ts
+++ b/packages/core/src/types/utility.ts
@@ -1,0 +1,5 @@
+
+/**
+ * Retrieve the type of fulfillment value of a Promise
+ */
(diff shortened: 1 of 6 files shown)


packages/core/src/types/utility.ts
```diff
@@ -0,0 +1,5 @@
+
+/**
+ * Retrieve the type of fulfillment value of a Promise
+ */
+export type PromisedValueOf<T extends Promise<any>> = T extends Promise<infer P> ? P : never
```
yorkie, line 5: May I ask you to move this to types.ts?

liangzr: Sure : )

Issue #163, opened 2020-05-03
cli: exit with code when some error happened.


Txiaozhe:
such as:
```js
export const start: StartHandler = async (filename: string) => {
  if (!filename) {
    spinner.fail('Please specify the config path');
    return;
  }

  if (!existsSync(filename)) {
    spinner.fail(`${filename} not exists`);
     return;
  }
```
return directly will make ci green although err was happened.

yorkie:
Yea, we should correct the exit code always, even though we don't throw an error. This causes the build on #137 is green on failures.

Txiaozhe:
This issue should be closed because of #164 and #162 has been merged:)

PR #164, opened 2020-05-03
cli: exit with code when failed.

Txiaozhe:
for #163

### DIFF
--- a/packages/cli/src/commands/board.ts
+++ b/packages/cli/src/commands/board.ts
@@ -11,7 +11,7 @@ export const board: CommandHandler = async () => {
   try {
     if (!fse.existsSync(path.join(process.cwd(), '.server'))) {
       spinner.fail('Please init the project firstly');
-      process.exit(1);
+      return process.exit(1);
     }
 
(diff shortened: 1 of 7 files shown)


packages/cli/src/commands/board.ts
```diff
@@ -11,7 +11,7 @@ export const board: CommandHandler = async () => {
   try {
     if (!fse.existsSync(path.join(process.cwd(), '.server'))) {
       spinner.fail('Please init the project firstly');
-      return;
+      process.exit(1);
```
yorkie, line 14:
```suggestion
      return process.exit(1);
```

packages/cli/src/commands/dataset.ts
```diff
@@ -16,7 +16,7 @@ import {
 export const dataset: DatasetCommandHandler = async ({ type }) => {
   if (!type) {
     console.log('Please provide a plugin type');
-    return;
+    process.exit(1);
```
yorkie, line 19: Ditto, return is also required.

Txiaozhe:
> `return` is also required to represent the logic flow.

Fixed.

PR #176, opened 2020-05-08
meta: add actions/labeler for auto labeling

yorkie:
This is going to help us to address the automatic label on Pull Request only.


utkobe:
Do we need to also add auto label for plugins

PR #177, opened 2020-05-08
core: add typings for runner

liangzr:
This PR improves the type of pipcook's runner as possible, and there still have some work TBD for `utils` or `types`

### DIFF
--- a/packages/core/src/runner/helper.ts
+++ b/packages/core/src/runner/helper.ts
@@ -127,20 +127,19 @@ export function linkComponents(components: PipcookComponentResult[]) {
  * those components which are not changed to success status should be failing ones
  * @param components 
  */
-export function assignFailures(components: PipcookComponentResult[]) {
-  components.forEach((component) => {
-    if (component.status === 'running') {
-      component.status = 'failure';
+export function markFailures(components: PipcookComponentResult[]): PipcookComponentResult[] {
+  return components.map(function markFailure(component) {
+    const nextComponent = { ...component };
+
+    if (nextComponent.status === 'running') {
+      nextComponent.status = 'failure';
     }
-    if (component.mergeComponents) {
-      component.mergeComponents.forEach((componentArray) => {
-        componentArray.forEach((component) => {
-          if (component.status === 'running') {
-            component.status = 'failure';
-          }
-        });
-      });
+
+    if (nextComponent.mergeComponents) {
+      nextComponent.mergeComponents = nextComponent.mergeComponents
+        .map((componentArray) => componentArray.map(markFailure));
     }
+
+    return nextComponent;
   });
 }
-
--- a/packages/core/src/runner/helper.ts
+++ b/packages/core/src/runner/helper.ts
@@ -4,20 +4,22 @@
 import * as path from 'path';
 import * as fs from 'fs-extra';
 import { from } from 'rxjs';
+import { flatMap } from 'rxjs/operators';
 
 import { PipcookRunner } from './index';
-import { PipcookComponentResult } from '../types/component';
-import { EvaluateError } from '../types/other';
-import { logCurrentExecution } from '../utils/logger';
-import { flatMap } from 'rxjs/operators';
 import { DATA, MODEL, EVALUATE, MODELTOSAVE } from '../constants/other';
+import { logCurrentExecution } from '../utils/logger';
+import { PipcookComponentResult, InsertParams, PipcookComponentOperator, PipcookComponentOutput } from '../types/component';
+import { EvaluateError, EvaluateResult, PipObject } from '../types/other';
+import { UniDataset } from '../types/data/common';
+import { UniModel } from '../types/model';
 
 /**
  * Retreive relative logs required to be stored.
  * @param pipcookRunner : The pipcookRunner object
  */
-export function getLog(pipcookRunner: PipcookRunner): any {
-  const result: any = {
+export function getLog(pipcookRunner: PipcookRunner) {
+  const result: PipObject = {
     ...pipcookRunner,
     latestModel: null,
     latestSampleData: null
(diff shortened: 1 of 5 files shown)


packages/core/src/runner/helper.ts
```diff
@@ -76,27 +78,25 @@ export function createPipeline(components: PipcookComponentResult[], self: Pipco
...
+  const insertParams: InsertParams = {
     pipelineId: self.pipelineId,
     modelDir: path.join(self.logDir, 'model'),
     dataDir: path.join(self.logDir, 'data')
   };
   const firstObservable = firstComponent.observer(null, self.latestModel, insertParams);
   self.updatedType = firstComponent.returnType;
 
-  const flatMapArray: any = [];
+  const flatMapArray: PipcookComponentOperator[] = [];
```
yorkie, line 89: How about using `reduce` function to remove the temp assignment?

packages/core/src/runner/helper.ts
```diff
@@ -128,13 +128,13 @@ export function linkComponents(components: PipcookComponentResult[]) {
...
-  components.forEach((component: PipcookComponentResult) => {
+  components.forEach((component) => {
     if (component.status === 'running') {
       component.status = 'failure';
     }
     if (component.mergeComponents) {
-      component.mergeComponents.forEach((componentArray: PipcookComponentResult[]) => {
-        componentArray.forEach((component: PipcookComponentResult) => {
+      component.mergeComponents.forEach((componentArray) => {
+        componentArray.forEach((component) => {
```
yorkie, line 137: Ditto, it's good to use map to goto next state to avoid update the current value.

packages/core/src/runner/helper.ts
```diff
@@ -37,29 +39,29 @@ export function getLog(pipcookRunner: PipcookRunner): any {
...
   case DATA:
-    self.latestSampleData = result;
+    self.latestSampleData = result as UniDataset;
     break;
   case MODEL:
-    self.latestModel = result;
+    self.latestModel = result as UniModel;
     break;
   case EVALUATE:
     console.log('evaluate result: ', result);
```
Txiaozhe, line 51: May console.log should be replaced?

packages/core/src/runner/helper.ts
```diff
@@ -37,29 +39,29 @@ export function getLog(pipcookRunner: PipcookRunner): any {
  * @param self: the target update runner
  * @param saveModelCallback
  */
-export async function assignLatestResult(updatedType: string, result: any, self: PipcookRunner, saveModelCallback?: Function) {
+export async function assignLatestResult(updatedType: string, result: PipcookComponentOutput, self: PipcookRunner, saveModelCallback?: Function) {
```
Txiaozhe, line 42: How about make 'updatedType' as a new type contains DATA, MODEL, EVALUATE etc. ?

Txiaozhe: IMO 'saveModelCallback' cloud be a specify type too.

packages/core/src/runner/helper.ts
yorkie: A TODO is to be added here, It is retained here because there is no ability to obtain the value of evaluate from the outside, so we need to use it to confirm whether the test passed, but we can also use `debug` instead, but I recommend that we complete it in another PR.

packages/core/src/runner/helper.ts
yorkie: IMO It depends on @liangzr, It's acceptable without the change and I'm +1 to refine the in a new PR :)

packages/core/src/runner/helper.ts
Txiaozhe: Approve:)

packages/core/src/runner/helper.ts
Txiaozhe: Great.

packages/core/src/types/component.ts
```diff
@@ -1,8 +1,8 @@
-import { Observable } from 'rxjs';
+import { Observable, OperatorFunction } from 'rxjs';
 import { PipcookPlugin } from './plugins';
```
Txiaozhe, line 2: I have modified this file, could you please pull and merge it?

packages/core/src/runner/helper.ts
liangzr: `reduce` is more elegant in coding, but less readable than a simple loop, It is more suitable when the result has the same type as the items

packages/core/src/runner/helper.ts
liangzr: Agree, assignment of the parameter may lead to unexpected side effects, should I optimize it in this PR?

packages/core/src/runner/helper.ts
liangzr: Actually I did want to change the type of `updatedType` to [const **literal** enums](https://www.typescriptlang.org/docs/handbook/enums.html#const-enums). Considering that its original type can at least work, I didn't modify it in this PR

packages/core/src/types/component.ts
liangzr: Sure!

packages/core/src/runner/helper.ts
yorkie: That's great if you can do it here.

liangzr: I have committed a new version includes:
- Convert `assignFailures` to pure function and renamed to `markFailures`, `assign` may imply a mutation of the object
- Mark failure status recursively. This will be more efficient and concise, but It introduces a pitfall that if `component` has a circular reference via `component.mergeComponents`, this function will cause a stack overflow. I am not sure if it can be used like this?

yorkie:
The labeler failed because:

> Secrets are not passed to workflows that are triggered by a pull request from a fork. Learn more.

utkobe:
auto labelling has been reverted. Please re-run jobs. Otherwise look good to me

yorkie:
> auto labelling has been reverted. Please re-run jobs. Otherwise look good to me

Never mind, we can merge it manually :)

PR #179, opened 2020-05-08
core: remove useless type.

Txiaozhe:


Txiaozhe:
'ResultType' is same with 'PluginTypeI', so remove it.

PR #182, opened 2020-05-09
examples: add gradient-boosting-regression

Heluwe:
Add sklearn gradient boosting regression example, see https://scikit-learn.org/stable/auto_examples/ensemble/plot_gradient_boosting_regression.html?highlight=feature_importance


yorkie:
Thank you @Heluwe, there are some style nits reported by GitHub, could you please fix them :)

PR #188, opened 2020-05-11
core: specify PipcookComponentResultStatus as type.

Txiaozhe:


yorkie:
@liangzr May I ask you to chime this in?

Txiaozhe:
> LGTM but the variable name `PipcookComponentResultStatus` is too long :(

But 'PipcookComponentResultStatus' is appropriate:(

liangzr:
@yorkie Totally!

Actually I'm still confused about the [`const string enums`] and [`string literal types`]. they're exactly the same after compiling.

TypeScript:
```ts
const enum Status {
  Running = 'running',
  Stopped = 'stopped'
}
type LiteralStatus = 'running' | 'stopped' 
const a: Status = Status.Running
const b: LiteralStatus = 'running'
```

JavaScript:
```js
const a = "running" /* Running */;
const b = 'running';
```

IMO, `const string enums` is more readable and opaque to the caller, `string literal types` is concise than an enum

[`const string enums`]: https://www.typescriptlang.org/docs/handbook/enums.html#const-enums
[`string literal types`]: https://www.typescriptlang.org/docs/handbook/literal-types.html

yorkie:
> IMO, const string enums is more readable and opaque to the caller, string literal types is concise than an enum

Yea, agreed with @liangzr's summary, so I tend to:

- enum could be used internally for better readable and abstraction.
- type alias could be used to our APIs, which could simplify the APIs.

Txiaozhe:
> > IMO, const string enums is more readable and opaque to the caller, string literal types is concise than an enum
> 
> Yea, agreed with @liangzr's summary, so I tend to:
> 
> * enum could be used internally for better readable and abstraction.
> * type alias could be used to our APIs, which could simplify the APIs.

But I am not very identification and IMO enum is more clear and definite. And enum will more convenient for users.

yorkie:
@Txiaozhe As I have commented at #189, using enum in TypeScript requires importing the Enum object at first, however the string-based type alias will not ask for, and if you are using a vscode, you will get choices when you are to assign a value which is declared in the type alias.

Txiaozhe:
> @Txiaozhe As I have commented at #189, using enum in TypeScript requires importing the Enum object at first, however the string-based type alias will not ask for, and if you are using a vscode, you will get choices when you are to assign a value which is declared in the type alias.

Sure:)

Issue #173, opened 2020-05-07
boa: use Node.js custom loader for better import-statement


yorkie:
See Node.js [ESM Experiment Loaders](https://nodejs.org/dist/latest-v14.x/docs/api/esm.html#esm_experimental_loaders), we could define a custom loader for Python modules just like:

```js
// node --experiment @pipcook/boa/esm/loader.mjs app.js
import { list, tuple } from 'py:builtins';
import { Image, ImageDraw } from 'py:PIL';
import { DBSCAN } from 'py:sklearn.cluster';
import { word2vec } from 'py:gensim.models';
```
Any good suggestion for the import name pattern?

Txiaozhe:
But this hook is experimental:( May we use it later?

yorkie:
> But this hook is experimental:( May we use it later?

The loader script is separated, so that's free to add.

Txiaozhe:
> > But this hook is experimental:( May we use it later?
> 
> The loader script is separated, so that's free to add.

Looks good~

rickyes:
Hi, I've been looking into trying to add `ESM` support for `Boa` lately and this might be something I can help with.

PR #191, opened 2020-05-14
boa: support esm

rickyes:
fixes: #173 

Support `export default` and `named export` import.

/cc @yorkie

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -11,7 +11,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        node_version: ['10', '12']
+        node_version: ['12', '14']
         os: [ubuntu-latest, macOS-latest]
     steps:
     - uses: actions/checkout@v1
--- a/docs/tutorials/want-to-use-python.md
+++ b/docs/tutorials/want-to-use-python.md
@@ -274,6 +274,40 @@ Returns the hash value of this object, internally it calls the CPython's [`PyObj
 
 Returns a corresponding primitive value for this object, see [`Symbol.toPrimitive` on MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol/toPrimitive) for more details.
 
+## Working with ECMAScript Modules
+> Requires Node.js >= `v12.11.1`
+
+Use Node.js custom loader for better import-statement.
+```js
+// app.mjs
+import { range, len } from 'py:builtins';
+import os from 'py:os';
+import {
+  array as NumpyArray,
+  int32 as NumpyInt32,
+} from 'py:numpy';
+
+console.log(os.getpid()); // prints the pid from python.
+
+const list = range(0, 10); // create a range array
+console.log(len(list)); // 10
+console.log(list[2]); // 2
+
+const arr = NumpyArray([1, 2, 3], NumpyInt32); // Create an array of int32 using ndarry constructor
+console.log(arr[0]); // 1
+```
+
+In `Node.js v14.x` you can specify only [`--experimental-loader`](https://nodejs.org/dist/latest-v14.x/docs/api/cli.html#cli_experimental_loader_module) to launch your application:
+
+```sh
+$ node --experimental-loader @pipcook/boa/esm/loader.mjs app.mjs
+```
+
+In Node.js version < `v14.x`, you also need to add the [`--experimental-modules`](https://nodejs.org/dist/latest-v14.x/docs/api/cli.html#cli_experimental_modules) option:
+```sh
+$ node --experimental-modules --experimental-loader @pipcook/boa/esm/loader.mjs app.mjs
+```
+
 ## Tests
 
 To run the full tests:
--- a/packages/boa/esm/loader.mjs
+++ b/packages/boa/esm/loader.mjs
@@ -1,0 +1,40 @@
+import boa from '../lib/index.js';
+const { dir } = boa.builtins();
+const extensionsPrefix = 'py:';
+const protocol = 'nodejs://boa/';
+
+export function resolve(specifier, context, defaultResolve) {
+  if (specifier.startsWith(extensionsPrefix)) {
+    return {
+      url: protocol + specifier.replace(extensionsPrefix, '')
+    };
+  }
+  return defaultResolve(specifier, context, defaultResolve);
+}
+
+export function getFormat(url, context, defaultGetFormat) {
+  // DynamicInstantiate hook triggered if boa protocol is matched
+  if (url.startsWith(protocol)) {
+    return {
+      format: 'dynamic',
+    }
+  }
+
+  // Other protocol are assigned to nodejs for internal judgment loading
+  return defaultGetFormat(url, context, defaultGetFormat);
+}
+
+export function dynamicInstantiate(url) {
+  const moduleInstance = boa.import(url.replace(protocol, ''));
+  // Get all the properties of the Python Object to construct named export
+  const moduleExports = dir(moduleInstance);
+  return {
+    exports: ['default', ...moduleExports],
+    execute: exports => {
+      for (let name of moduleExports) {
+        exports[name].set(moduleInstance[name]);
+      }
+      exports.default.set(moduleInstance);
+    }
+  };
+}
--- a/packages/boa/package.json
+++ b/packages/boa/package.json
@@ -6,7 +6,7 @@
   "scripts": {
     "preinstall": "node tools/install-python.js && node tools/install-requirements.js",
     "postinstall": "npm run build",
-    "test": "tap ./tests/**/*.js",
+    "test": "tap --no-cov ./tests/**/*.js",
     "lint": "eslint . && ./clang-format.py",
     "build": "node-gyp configure && node-gyp build",
     "pretest": "npm run lint",
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,7 @@ node_js:
   - '8'
   - '10'
   - '12'
+  - '14'
 before_install:
   - npm i npminstall -g
 install:
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -3,6 +3,7 @@ environment:
     - nodejs_version: '8'
     - nodejs_version: '10'
     - nodejs_version: '12'
+    - nodejs_version: '14'
 
 install:
   - ps: Install-Product node $env:nodejs_version
--- a/docs/zh-cn/tutorials/want-to-use-python.md
+++ b/docs/zh-cn/tutorials/want-to-use-python.md
@@ -274,6 +274,40 @@ Returns the hash value of this object, internally it calls the CPython's [`PyObj
 
 Returns a corresponding primitive value for this object, see [`Symbol.toPrimitive` on MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol/toPrimitive) for more details.
 
+## Working with ECMAScript Modules
+> Requires Node.js >= `v12.11.1`
+
+Use Node.js custom loader for better import-statement.
+```js
+// app.mjs
+import { range, len } from 'py:builtins';
+import os from 'py:os';
+import {
+  array as NumpyArray,
+  int32 as NumpyInt32,
+} from 'py:numpy';
+
+console.log(os.getpid()); // prints the pid from python.
+
+const list = range(0, 10); // create a range array
+console.log(len(list)); // 10
+console.log(list[2]); // 2
+
+const arr = NumpyArray([1, 2, 3], NumpyInt32); // Create an array of int32 using ndarry constructor
+console.log(arr[0]); // 1
+```
+
+In `Node.js v14.x` you can specify only [`--experimental-loader`](https://nodejs.org/dist/latest-v14.x/docs/api/cli.html#cli_experimental_loader_module) to launch your application:
+
+```sh
+$ node --experimental-loader @pipcook/boa/esm/loader.mjs app.mjs
+```
+
+In Node.js version < `v14.x`, you also need to add the [`--experimental-modules`](https://nodejs.org/dist/latest-v14.x/docs/api/cli.html#cli_experimental_modules) option:
+```sh
+$ node --experimental-modules --experimental-loader @pipcook/boa/esm/loader.mjs app.mjs
+```
+
 ## Tests
 
 To run the full tests:
--- a/packages/boa/tests/es-module-loaders/common.js
+++ b/packages/boa/tests/es-module-loaders/common.js
@@ -1,0 +1,50 @@
+/* eslint-disable no-process-exit */
+
+'use strict';
+
+const path = require('path');
+const { test } = require('tap');
+const { spawnSync } = require('child_process');
+
+function getAbsolutePath(relativePath) {
+  return path.join(__dirname, relativePath);
+}
+
+const FLAG = '--experimental-loader';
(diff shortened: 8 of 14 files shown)


packages/boa/esm/loader.mjs
```diff
@@ -0,0 +1,39 @@
+import Boa from '../lib/index.js';
```
yorkie, line 1:
```suggestion
import Boa from '../';
```

packages/boa/esm/loader.mjs
```diff
@@ -0,0 +1,39 @@
...
+}
+
+export function dynamicInstantiate(url) {
+  const moduleInstance = Boa.import(url.replace(passBaseUrl, ''));
+  const moduleExports = ['default', ...Object.keys(moduleInstance)];
+  return {
+    exports: moduleExports,
+    execute: exports => {
+      for (let name of moduleExports)
+        exports[name].set(moduleInstance[name]);
```
yorkie, line 35:
```suggestion
      for (let name of moduleExports) {
        exports[name].set(moduleInstance[name]);
      }
```

packages/boa/esm/loader.mjs
```diff
@@ -0,0 +1,39 @@
...
+      exports.default.set(moduleInstance);
+    }
+  };
+}
```
yorkie, line 39: EOF

yorkie:
Thank you @rickyes I have commented some, and unit test is also required :)

packages/boa/esm/loader.mjs
rickyes: Doesn't seem to automatically look for `package.json`.

packages/boa/esm/loader.mjs
```diff
@@ -0,0 +1,39 @@
...
+  }
+
+  return {
+    format: format || 'module'
+  };
+}
+
+export function dynamicInstantiate(url) {
+  const moduleInstance = Boa.import(url.replace(passBaseUrl, ''));
+  const moduleExports = ['default', ...Object.keys(moduleInstance)];
```
Txiaozhe, line 30: Missing space for '[]' :)

packages/boa/esm/loader.mjs
rickyes: fixed

packages/boa/esm/loader.mjs
rickyes: fixed. Maybe we can add some eslint rules.

rickyes:
> Thank you @rickyes I have commented some, and unit test is also required :)

Of course, when the deconstruction import is supported, unit tests will be added together.

.github/workflows/build.yml
```diff
@@ -11,7 +11,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        node_version: ['10', '12']
+        node_version: ['10', '12', '14']
```
yorkie, line 14: Why adding this node14 support?

rickyes: `--experimental-loader` option is not supported before node 14.

packages/boa/esm/loader.mjs
```diff
@@ -0,0 +1,40 @@
...
+  return {
+    format: format || 'module'
+  };
+}
+
+export function dynamicInstantiate(url) {
+  const moduleInstance = boa.import(url.replace(passBaseUrl, ''));
+  const moduleExports = [ 'default', ...Object.keys(moduleInstance) ];
+  return {
+    exports: moduleExports,
```
rickyes, line 32: Need to pass named, can't think of a way to get py module's properties. @yorkie Any good suggestions？

yorkie: Boa fetches the property object dynamically, can esm loader support absolute dynamic way?

yorkie: As for fetching the props statically, you can try with `obj.__dict__`, which returns the properties of the given object.

.github/workflows/build.yml
yorkie: See https://nodejs.org/dist/latest-v12.x/docs/api/esm.html#esm_experimental_loaders, it seems working at node12, too.

packages/boa/esm/loader.mjs
yorkie: Weird, we have defined the `main` as "lib/index.js" at https://github.com/alibaba/pipcook/blob/master/packages/boa/package.json#L5.

.github/workflows/build.yml
rickyes: This option is enabled in node 12 and needs to be used in conjunction with `-experimental-modules', but `--experimental-modules` is deprecated in node 14.

rickyes: Maybe we can add backwards compatible unit tests.

yorkie: I think we can drop the node10, and use different flags for 12 and 14.

rickyes: fixed

packages/boa/tests/fixtures/es-module-loaders/common.js
```diff
@@ -0,0 +1,35 @@
...
+}
+
+function checkExec(t, appPath) {
+  const options = { encoding: 'utf8', stdio: 'inherit' };
+  const args = [FLAG, PATH_ESM_LOADER];
+  // See https://github.com/nodejs/node/pull/31974
+  if (process.version < 'v14.0.0') {
+    args.push('--experimental-modules');
+  }
+  const result = spawnSync(process.execPath, args.concat(appPath), options);
```
yorkie, line 27: This might be unneeded, it's recommended that adding a new script at https://github.com/alibaba/pipcook/blob/master/packages/boa/package.json#L9.

packages/boa/esm/loader.mjs
rickyes: fixed

rickyes:
The `named export` is already supported, please review. @yorkie @Txiaozhe

packages/boa/tests/fixtures/es-module-loaders/test-esm-loader-export-default.mjs
```diff
@@ -0,0 +1,9 @@
+import os from 'py:os';
+import { len, range } from 'py:builtins';
+import { strictEqual } from 'assert';
+import { add, subtract } from './test-esm-loader-math.mjs';
```
yorkie, line 4: Is this necessary to test for?

packages/boa/src/core/object.cc
```diff
@@ -324,6 +325,12 @@ Napi::Value PythonObject::SetItem(const CallbackInfo &info) {
   return Number::New(info.Env(), r);
 }
 
+Napi::Value PythonObject::Dir(const CallbackInfo &info) {
+  PyObject *attrs = PyObject_Dir(_self.ptr());
```
yorkie, line 329: As the Python documentation said:

> This is equivalent to the Python expression dir(o).

Can we implement this in pure JavaScript?

packages/boa/tests/fixtures/es-module-loaders/test-esm-loader-export-default.mjs
rickyes: Yes, I think it's in the change range too.

packages/boa/src/core/object.cc
rickyes: May not be available in pure JavaScript at the moment, the dynamicInstantiate hook cannot pass proxy objects for dynamic import.

rickyes: I will look into the possibility of implementing dynamic import in the node core as our next step of optimization.

rickyes:
> I have some comments for your changes, and we also need more tests, which needs cover builtins, base package and numpy.

Okay, I'll supplement the base library test.

packages/boa/esm/loader.mjs
```diff
@@ -0,0 +1,40 @@
...
+  }
+
+  return {
+    format: format || 'module'
+  };
+}
+
+export function dynamicInstantiate(url) {
+  const moduleInstance = boa.import(url.replace(passBaseUrl, ''));
+  const moduleExports = moduleInstance.__dir__();
```
yorkie, line 30:
```js
const { dir } = boa.builtins();
const moduleExports = dir(moduleInstance);
```

rickyes:
I've made some changes, please review them.

docs/tutorials/want-to-use-python.md
```diff
@@ -274,6 +274,34 @@ Returns the hash value of this object, internally it calls the CPython's [`PyObj
 
 Returns a corresponding primitive value for this object, see [`Symbol.toPrimitive` on MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol/toPrimitive) for more details.
 
+#### With ES-Modules
+> Requires Node.js >= `v12.11.1`
+
+use Node.js custom loader for better import-statement.
```
yorkie, line 280:
```suggestion
Use Node.js custom loader for better import-statement.
```

docs/tutorials/want-to-use-python.md
```diff
@@ -274,6 +274,34 @@ Returns the hash value of this object, internally it calls the CPython's [`PyObj
...
+console.log(os.getpid()); // prints the pid from python.
+
+const list = range(0, 10); // create a range array
+console.log(len(list)); // 10
+console.log(list[2]); // 2
+```
+
+In `Node.js v14.x` you can specify only [`--experimental-loader`](https://nodejs.org/dist/latest-v14.x/docs/api/cli.html#cli_experimental_loader_module) to launch your application:
+
+``` sh
```
yorkie, line 296:
```suggestion
```sh
```

docs/tutorials/want-to-use-python.md
```diff
@@ -274,6 +274,34 @@ Returns the hash value of this object, internally it calls the CPython's [`PyObj
 
 Returns a corresponding primitive value for this object, see [`Symbol.toPrimitive` on MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol/toPrimitive) for more details.
 
+#### With ES-Modules
+> Requires Node.js >= `v12.11.1`
+
+use Node.js custom loader for better import-statement.
+``` js
```
yorkie, line 281:
```suggestion
```js
```

docs/tutorials/want-to-use-python.md
```diff
@@ -274,6 +274,34 @@ Returns the hash value of this object, internally it calls the CPython's [`PyObj
...
+// app.mjs
+import { equal, ok } from 'assert';
+import { range, len } from 'py:builtins';
+import os from 'py:os';
+
+console.log(os.getpid()); // prints the pid from python.
+
+const list = range(0, 10); // create a range array
+console.log(len(list)); // 10
+console.log(list[2]); // 2
```
yorkie, line 291: IMO it's good to add a numpy example here.

packages/boa/package.json
```diff
@@ -6,7 +6,7 @@
   "scripts": {
     "preinstall": "node tools/install-python.js && node tools/install-requirements.js",
     "postinstall": "npm run build",
-    "test": "tap ./tests/**/*.js",
+    "test": "nyc tap ./tests/**/*.js",
```
yorkie, line 9: Please keep using the tap only, we are still not ready for code coverage with nyc :(

rickyes: No need to coverage  code? then just using `tap --no-cov` will suffice, if using `nyc 14` will trigger an error, please see [--experimental-loader does not work](https://github.com/tapjs/node-tap/issues/673#issuecomment-629813258)

docs/tutorials/want-to-use-python.md
rickyes: fixed

packages/boa/package.json
yorkie: Pipcook children projects are maintained by lerna, and we have to consider about code coverage from the whole projects.

And this PR is to add the esm support only, so IMO this change on build stuffs are not proper here :)

docs/tutorials/want-to-use-python.md
```diff
@@ -274,6 +274,40 @@ Returns the hash value of this object, internally it calls the CPython's [`PyObj
 
 Returns a corresponding primitive value for this object, see [`Symbol.toPrimitive` on MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol/toPrimitive) for more details.
 
+#### With ES-Modules
```
yorkie, line 277:
```suggestion
## Working with ECMAScript Modules
```
I read the entire document and feel that it should be placed in the level 2 :)

docs/tutorials/want-to-use-python.md
```diff
@@ -274,6 +274,40 @@ Returns the hash value of this object, internally it calls the CPython's [`PyObj
...
+console.log(arr[0]); // 1
+```
+
+In `Node.js v14.x` you can specify only [`--experimental-loader`](https://nodejs.org/dist/latest-v14.x/docs/api/cli.html#cli_experimental_loader_module) to launch your application:
+
+```sh
+$ node --experimental-loader @pipcook/boa/esm/loader.mjs app.mjs
+```
+
+In Node.js Version < `v14.x`, you also need to add the [`--experimental-modules`](https://nodejs.org/dist/latest-v14.x/docs/api/cli.html#cli_experimental_modules) option:
```
yorkie, line 306:
```suggestion
In Node.js version < `v14.x`, you also need to add the [`--experimental-modules`](https://nodejs.org/dist/latest-v14.x/docs/api/cli.html#cli_experimental_modules) option:
```

packages/boa/package.json
rickyes: Ok, but if left as is the CI run will report an error, I'm not sure if the new PR is needed to remove the code coverage first.

docs/tutorials/want-to-use-python.md
rickyes: done

packages/boa/package.json
yorkie: We can first see what errors are encountered, and it seems that we don't add code coverage.

yorkie:
@rickyes May I ask you to revert the code coverage? It seems that LGTM except that :)

packages/boa/package.json
rickyes: See https://github.com/tapjs/node-tap/issues/673

yorkie:
```
internal/modules/run_main.js:54
    internalBinding('errors').triggerUncaughtException(
                              ^

TypeError [ERR_UNKNOWN_FILE_EXTENSION]: Unknown file extension "" for /home/runner/.node-spawn-wrap-5008-e0ba0a808e26/node
    at Loader.defaultGetFormat [as _getFormat] (internal/modules/esm/get_format.js:65:15)
    at Loader.getFormat (internal/modules/esm/loader.js:113:42)
    at Loader.getModuleJob (internal/modules/esm/loader.js:244:31)
    at processTicksAndRejections (internal/process/task_queues.js:97:5)
    at async Loader.import (internal/modules/esm/loader.js:178:17)
    at async internal/process/esm_loader.js:55:9 {
  code: 'ERR_UNKNOWN_FILE_EXTENSION'
}
```
It seems the above error occurred.

packages/boa/package.json
yorkie: Thank you, after reading @coreyfarrell's comment, IMO we have to add `--no-cov` here.

rickyes: fixed

PR #192, opened 2020-05-15
chore: add .editorconfig configuration to automatically add EOF

rickyes:


.editorconfig
```diff
@@ -0,0 +1,5 @@
+  
+root = true
+
+[*]
+insert_final_newline = true
```
yorkie, line 5: The EOF is missing :p

rickyes: hhh, fixed :)

PR #195, opened 2020-05-21
move utils from pipcook-core to single package

utkobe:
This is to decouple pipcook-core which contains types definition of pipcook with any utils functions that dependent on nodejs module. So that pipcook-core can also be used in browser (pipboard).


packages/cli/package.json
```diff
@@ -37,7 +37,8 @@
     "request": "^2.88.2",
     "semver": "^6.3.0",
     "ts-node": "^8.9.1",
-    "yaml": "^1.7.2"
+    "yaml": "^1.7.2",
+    "@pipcook/pipcook-utils": "^0.5.0"
```
yorkie, line 41: Sort the list in alphabeta :)

packages/cli/src/commands/serve.ts
```diff
@@ -2,13 +2,12 @@ import Fastify from 'fastify';
 import path from 'path';
 import ora from 'ora';
 import childProcess from 'child_process';
-import { constants } from '@pipcook/pipcook-core';
+import { PIPCOOK_LOGS } from '@pipcook/pipcook-utils';
```
yorkie, line 5: Constants should be in core?

packages/cli/src/config.ts
```diff
@@ -1,3 +1,6 @@
+import * as path from "path";
```
yorkie, line 1: Never used.

packages/cli/src/config.ts
```diff
@@ -1,3 +1,6 @@
...
+import * as os from 'os';
```
yorkie, line 2: Ditto.

packages/daemon/package.json
```diff
@@ -12,6 +12,7 @@
     "uuid": "^8.0.0",
     "uuid-validate": "^0.0.3",
     "@pipcook/pipcook-core": "^0.5.16",
+    "@pipcook/pipcook-utils": "^0.5.0",
```
yorkie, line 15:
```suggestion
    "@pipcook/utils": "^0.5.0",
```

packages/plugins/data-collect/image-classification-data-collect/package.json
```diff
@@ -21,7 +21,8 @@
     "fs-extra": "^8.1.0",
     "glob": "^7.1.6",
     "glob-promise": "^3.4.0",
-    "uuid": "^7.0.2"
+    "uuid": "^7.0.2",
+    "@pipcook/pipcook-utils": "^0.5.0"
```
yorkie, line 25: IMO the plugin won't depend on core.

packages/utils/package.json
```diff
@@ -0,0 +1,44 @@
+{
+  "name": "@pipcook/pipcook-utils",
```
yorkie, line 2:
```suggestion
  "name": "@pipcook/utils",
```

packages/utils/src/constants.ts
```diff
@@ -0,0 +1,7 @@
+import * as os from 'os';
+import * as path from 'path';
+
+export const PIPCOOK_HOME_PATH = path.join(os.homedir(), '.pipcook');
+export const PIPCOOK_PLUGINS = path.join(PIPCOOK_HOME_PATH, 'plugins');
+export const PIPCOOK_LOGS = path.join(PIPCOOK_HOME_PATH, 'logs');
+export const PIPCOOK_DEPENDENCIES = path.join(PIPCOOK_HOME_PATH, 'dependencies');
```
Txiaozhe, line 7: How about aggregate these paths to a named constant such as 'PIPCOOK_PATH'?

packages/cli/src/commands/serve.ts
utkobe: These constants are path related which requires path module. So I move them into utils

packages/plugins/data-collect/image-classification-data-collect/package.json
utkobe: What's the reason. Core provides some definitions of types of plugins

packages/utils/src/constants.ts
utkobe: Good idea. Fixed

packages/utils/package.json
utkobe: fixed

packages/daemon/package.json
utkobe: fixed

packages/cli/src/config.ts
utkobe: fixed

packages/cli/src/config.ts
utkobe: fixed

packages/cli/package.json
utkobe: fixed

packages/plugins/data-collect/image-classification-data-collect/package.json
yorkie: IMO, the core will be our internal lib, and we need to keep fucking simple for plugin developers.

packages/cli/src/commands/serve.ts
yorkie: As I have thought at https://github.com/alibaba/pipcook/pull/195#discussion_r429023922, if the core is used internally, it depends on if we need to provide these paths to plugin developers.

PR #196, opened 2020-05-21
boa: Modify the Python path protocol to py:

rickyes:
With simple `py:` protocol to transfer between hook.


yorkie:
I remember that Node.js will check the protocol in node and file?

rickyes:
> I remember that Node.js will check the protocol in node and file?

@yorkie Sorry, I read the wrong one before, and directly returning the url will not check the url, only the default `defaultResolve` function will check the URL format. 
See https://github.com/nodejs/node/blob/master/lib/internal/modules/esm/loader.js#L97-L109

PR #201, opened 2020-06-01
cli: add beta options

yorkie:


utkobe:
Probably we also need to add a beta option in pipcook run so that the costa will also install plugins with beta versions? Or can we just add version in config json to specify the version of plugins?

yorkie:
> Probably we also need to add a beta option in the pipcook run so that the costa will also install plugins with beta versions? Or can we just add a version in config JSON to specify the version of plugins?

See https://github.com/alibaba/pipcook/pull/202, I have supported the beta version there :)

yorkie:
@utkobe and IMO this PR doesn't depend on #202, to use beta plugins, developers should specify the version in the pipeline config after #202 is merged :)

utkobe:
LGTM:)

Issue #200, opened 2020-06-01
pipeline: support start a pipeline from a remote config url


yorkie:
With #150, we have achieved an easy way to start from a local pipeline config, however we need to move forward to better way that is start a pipeline from a url.

You can look at [the source](https://github.com/alibaba/pipcook/blob/master/packages/daemon/src/app/controller/job.ts#L32)  to get started.

yorkie:
This is completed at #203.

PR #203, opened 2020-06-02
daemon: support config url

FeelyChau:
#200 
Support config-url.
You can run as:
```shell
$ pipcook run http://host/config.json
```


yorkie:
@FeelyChau Thanks for this contribution, and just have a double check before you submitting a pull request next time:

- run `npm run lint` locally to make sure the lint pass.
- it's recommended that using `fix/gh-{id}` as the branch name if you are going to fix issue on GitHub.

packages/cli/src/utils.ts
```diff
@@ -24,3 +24,6 @@ export function tail(id: string, name: string): ChildProcess {
     }
   );
 }
+export function isUrl(url: string) {
+  return /^https?:\/\/.+/.test(url)
```
yorkie, line 28: How about Node.js `url.parse()`?

packages/daemon/src/runner/helper.ts
```diff
@@ -13,8 +14,13 @@ import {
 
 const { PLUGINS, PIPCOOK_LOGS } = constants;
 
-export async function parseConfig(configPath: string, generateId = true): Promise<PipelineDB> {
-  const configJson: RunConfigI = await fs.readJson(configPath);
+export async function parseConfig(config: string, generateId = true): Promise<PipelineDB> {
```
yorkie, line 17: The function signature should not be changed :)

packages/daemon/src/runner/helper.ts
```diff
@@ -13,8 +14,13 @@ import {
...
+  let configJson: RunConfigI = null;
+  if (/^https?:\/\/.+/.test(config)) {
+    configJson = JSON.parse(await request(config));
```
yorkie, line 20: Maybe we need a `loadJson()` at utils?

packages/cli/src/actions/start.ts
```diff
@@ -15,10 +15,12 @@ const start: StartHandler = async (filename: string, opts: any) => {
     spinner.fail('Please specify the config path');
     return process.exit(1);
   }
-  filename = path.isAbsolute(filename) ? filename : path.join(process.cwd(), filename);
-  if (!existsSync(filename)) {
-    spinner.fail(`${filename} not exists`);
-    return process.exit(1);
+  if (!isUrl(filename)) {
```
yorkie, line 18: Why do we add this check?

packages/cli/src/utils.ts
FeelyChau: This PR supports http and https only, `url.parse()` is out of the protocols.

packages/cli/src/actions/start.ts
FeelyChau: If not, the file url will be joined the file path prefix, we should skip this step if it's a url.

yorkie: How about moving the existence check to daemon? So here is to do convert filename to file-protocol URI, and the daemon API might handle the URI only.

packages/cli/src/utils.ts
yorkie: `url` can handle http/https, try `urlobj.protocol`.

packages/daemon/src/runner/helper.ts
FeelyChau: But the meaning of `configPath` has actually changed.

packages/daemon/src/runner/helper.ts
FeelyChau: Use `url.parse()` to get the url protocol.

packages/cli/src/actions/start.ts
FeelyChau: > How about moving the existence check to daemon? So here is to do convert filename to file-protocol URI, and the daemon API might handle the URI only.

It looks more reasonable.

packages/daemon/package.json
```diff
@@ -8,15 +8,16 @@
...
     "chalk": "^3.0.0",
     "cls-hooked": "^4.2.2",
+    "debug": "^4.1.1",
     "egg": "^2.26.1",
     "fs-extra": "^8.1.0",
     "glob": "^7.1.6",
     "glob-promise": "^3.4.0",
     "jimp": "^0.10.0",
     "midway": "^1.19.0",
+    "request-promise": "^4.2.5",
```
yorkie, line 20: Missing `@types/request-promise`.

FeelyChau: Done.

.github/workflows/cli.yml
```diff
@@ -57,9 +57,12 @@ jobs:
       run: |
         ./packages/cli/dist/bin/pipcook init
         ./packages/cli/dist/bin/pipcook daemon start
-    - name: pipcook run
+    - name: pipcook run local path
       run: |
         ./packages/cli/dist/bin/pipcook run ./test/pipelines/text-bayes-classification.json --verbose
+    - name: pipcook run url
```
yorkie, line 63:
```suggestion
    - name: pipcook run from url
```

.github/workflows/cli.yml
```diff
@@ -57,9 +57,12 @@ jobs:
       run: |
         ./packages/cli/dist/bin/pipcook init
         ./packages/cli/dist/bin/pipcook daemon start
-    - name: pipcook run
+    - name: pipcook run local path
```
yorkie, line 60:
```suggestion
    - name: pipcook run from local filename
```

packages/daemon/src/runner/helper.ts
yorkie: It acts as the pathname for pipeline config.

packages/cli/src/actions/start.ts
```diff
@@ -15,10 +16,18 @@ const start: StartHandler = async (filename: string, opts: any) => {
...
-    spinner.fail(`${filename} not exists`);
-    return process.exit(1);
+  const urlObj = url.parse(filename);
+  if (urlObj.protocol) {
+    if (urlObj.protocol !== 'http:' && urlObj.protocol !== 'https:') {
+      spinner.fail(`protocol ${urlObj.protocol} is not supported`);
+      return process.exit(1);
+    }
+  } else {
+    filename = path.isAbsolute(filename) ? filename : path.join(process.cwd(), filename);
```
yorkie, line 26: We could convert it to the file-protocol.

packages/daemon/src/runner/helper.ts
```diff
@@ -13,8 +15,19 @@ import {
...
-  const configJson: RunConfigI = await fs.readJson(configPath);
+export async function parseConfig(config: string, generateId = true): Promise<PipelineDB> {
+  let configJson: RunConfigI = null;
+  let urlObj = null;
+  urlObj = url.parse(config);
+  if (urlObj.protocol) {
+    if (urlObj.protocol === 'http:' || urlObj.protocol === 'https:') {
+      configJson = JSON.parse(await request(config));
+    } else {
+      throw new Error(`protocol ${urlObj.protocol} is not supported`);
```
yorkie, line 26:
```suggestion
      throw new TypeError(`protocol ${urlObj.protocol} is not supported`);
```

packages/cli/src/actions/start.ts
FeelyChau: How about we refactoring this into file-upload.

yorkie: You added the extra check at daemon, so I think it's required at this PR :)

FeelyChau:
> Besides, IMO we can add a new step to run a pipeline remotely in our [`workflows/cli.yml`](https://github.com/alibaba/pipcook/blob/master/.github/workflows/cli.yml#L62) :)

Done.

packages/cli/src/actions/start.ts
```diff
@@ -15,10 +16,20 @@ const start: StartHandler = async (filename: string, opts: any) => {
...
+    if (!existsSync(filename)) {
+      spinner.fail(`${filename} not exists`);
+      return process.exit(1);
+    }
+    urlObj = url.parse(`file://${filename}`);
+    filename = urlObj.href;
```
yorkie, line 32:
```js
const urlObj = url.parse(filename);
if (urlObj.protocol == null) {
  // FIXME(feely): file by default if protocol is null
  filename = path.isAbsolute(filename) ? filename : path.join(process.cwd(), filename);
  // check the filename existence
  if (!existsSync(filename)) {
    return process.exit(1);
  }
  filename = `file://${filename}`;
} else if (urlObj.protocol.indexOf([ 'http:', 'https:' ]) === -1) {
  return process.exit(1);
}
```

packages/daemon/src/runner/helper.ts
```diff
@@ -13,8 +15,21 @@ import {
 
 const { PLUGINS, PIPCOOK_LOGS } = constants;
 
-export async function parseConfig(configPath: string, generateId = true): Promise<PipelineDB> {
-  const configJson: RunConfigI = await fs.readJson(configPath);
+export async function parseConfig(config: string, generateId = true): Promise<PipelineDB> {
+  let configJson: RunConfigI = null;
+  let urlObj = null;
+  urlObj = url.parse(config);
+  if (urlObj.protocol) {
```
yorkie, line 22: We can throw it firstly just like:

```js
if (urlObj.protocol == null) {
  throw TypeError(...);
}
```

run_pipeline.sh
```diff
@@ -1,3 +1,3 @@
 #!/usr/bin/env sh
 
-./packages/cli/dist/bin/pipcook run ./test/pipelines/$1.json --verbose
+./packages/cli/dist/bin/pipcook run "./test/pipelines/$1.json" --verbose
```
yorkie, line 3: Which case do you make this change?

packages/daemon/src/runner/helper.ts
```diff
@@ -14,7 +16,20 @@ import {
 const { PLUGINS, PIPCOOK_LOGS } = constants;
 
 export async function parseConfig(configPath: string, generateId = true): Promise<PipelineDB> {
-  const configJson: RunConfigI = await fs.readJson(configPath);
+  let configJson: RunConfigI = null;
```
yorkie, line 19:
```suggestion
  let configJson: RunConfigI;
```

packages/daemon/src/runner/helper.ts
FeelyChau: OK.

packages/daemon/src/runner/helper.ts
```diff
@@ -14,7 +16,20 @@ import {
...
+  let urlObj = null;
```
yorkie, line 20:
```suggestion
  let urlObj = url.parse(configPath);
```

run_pipeline.sh
FeelyChau: Filename with space.

packages/cli/src/actions/start.ts
FeelyChau: `filename` should be encoded if we use file-protocol.

packages/daemon/src/runner/helper.ts
FeelyChau: Done.

packages/cli/src/actions/start.ts
FeelyChau: Done.

packages/daemon/src/runner/helper.ts
FeelyChau: Done.

packages/daemon/src/runner/helper.ts
FeelyChau: Done.

packages/cli/src/actions/start.ts
```diff
@@ -15,9 +16,19 @@ const start: StartHandler = async (filename: string, opts: any) => {
...
+  let urlObj = url.parse(filename);
+  // file default if the protocol is null
+  if (urlObj.protocol === null) {
+    filename = path.isAbsolute(filename) ? filename : path.join(process.cwd(), filename);
+    // check the filename existence
+    if (!existsSync(filename)) {
+      spinner.fail(`${filename} not exists`);
+      return process.exit(1);
+    } else {
+      filename = url.parse(`file://${filename}`).href;
```
yorkie, line 28: IMO, why not using `file://${filename}` directly?

packages/daemon/.github/workflows/nodejs.yml
```diff
@@ -0,0 +1,31 @@
+# This workflow will do a clean install of node dependencies, build the source code and run tests across different versions of node
```
yorkie, line 1: What's this new file for?

run_pipeline.sh
yorkie: Do we have pipeline filename with spaces?

FeelyChau: It's depended on the user.

packages/cli/src/actions/start.ts
FeelyChau: The string `file://${filename}` could be an unsafe URI.
Look at:
https://tools.ietf.org/html/rfc3986#section-2

packages/daemon/.github/workflows/nodejs.yml
FeelyChau: Unused file, removed.

utkobe:
LGTM :)

PR #211, opened 2020-06-05
daemon: cleanup if installing a package is failed

yorkie:
We should uninstall the plugin if the installation is failed finally.


utkobe:
LGTM :)

PR #212, opened 2020-06-05
boa: share the cpython instance globally

yorkie:
The CPython is a single instance, so we need to keep consistent with CPython by storing its instance globally.

This fixes the issue "python interrupter is already running" when loading the boa in different plugins.

### DIFF
--- a/packages/boa/lib/index.js
+++ b/packages/boa/lib/index.js
@@ -10,8 +10,13 @@ const DelegatorLoader = require('./delegator-loader');
 // internal symbols
 const IterIdxForSeqSymbol = Symbol('The iteration index for sequence');
 
-// create the instance
-let pyInst = new native.Python(process.argv.slice(1));
+// create the global-scoped instance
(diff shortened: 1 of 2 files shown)


utkobe:
LGTM except CI problem

yorkie:
Restart a new, and failed by a timeout.

packages/boa/lib/index.js
```diff
@@ -10,8 +10,13 @@ const DelegatorLoader = require('./delegator-loader');
 // internal symbols
 const IterIdxForSeqSymbol = Symbol('The iteration index for sequence');
 
-// create the instance
-let pyInst = new native.Python(process.argv.slice(1));
+// create the global-scoped instance
+let pyInst = global.__pipcook_boa_pyinst__;
+if (pyInst == null) {
```
rickyes, line 15:
```suggestion
if (pyInst === null || pyInst === undefined) {
```
Maybe we could use full equals.

PR #219, opened 2020-06-06
doc: update the readme

yorkie:


README.md
```diff
@@ -34,62 +34,77 @@ push the whole area forward with the machine learning engineering. For this reas
...
+$ pipcook init --tuna
 ```
 
-Occuring the download problems? We use [tuna](https://mirror.tuna.tsinghua.edu.cn/help/pypi/) mirror to address this issue:
+Then run a pipeline:
 
-```sh
-$ pipcook init --tuna
+```shell
+$ pipcook run https://raw.githubusercontent.com/alibaba/pipcook/master/example/pipelines/text-bayes-classification.json
```
utkobe, line 99: Suggest to give out --verbose version of this command.  Probably also give another command for --tuna args.

yorkie: Should be the `--verbose` option set by default?

rickycao-qy: I think so for new users

yorkie: So shall we keep this here and change it at CLI?

rickycao-qy: Good idea

PR #222, opened 2020-06-07
cli: set spinner output stream to stdout

FeelyChau:
As default the stream of `ora` (spinner) is stderr,  refactor to stdout for pipeline operation of shell.


packages/cli/src/utils.ts
```diff
@@ -1,5 +1,6 @@
 import * as os from 'os';
 import { exec, spawn, ChildProcess, ExecOptions, ExecException } from 'child_process';
+import realOra from 'ora';
```
yorkie, line 3:
```suggestion
import * as realOra from 'ora';
```

PR #228, opened 2020-06-10
upgrade docker file to version 1.1

wordcount:
update Dockerfile，use nodejs-12 as base


rickycao-qy:
LGTM

wordcount:
> May I ask you to not change the Dockerfile path? This is broken with https://github.com/alibaba/pipcook/blob/master/docs/INSTALL.md#install-via-docker.

Normally docker will add all file as it's build context, this will add build time and may cause other problems.
>[reference of docker build command](https://docs.docker.com/engine/reference/commandline/build/)
In most cases, it’s best to put each Dockerfile in an empty directory. Then, add to that directory only the files needed for building the Dockerfile.

yorkie:
Reasonable @wordcount, May I ask you to update the documentation, too?

wordcount:
> Reasonable @wordcount, May I ask you to update the documentation, too?

done

wordcount:
> LGTM even though the markdown links might be 404.

where

PR #230, opened 2020-06-11
boa: Compiling-CPython-to-WebAssembly

rickyes:
Compiling CPython to WebAssembly, easy to support boa on more platforms in the future. 

work in progress.


yorkie:
Is this from third-party https://github.com/rickyes/pipcook/tree/Compiling-CPython-to-WebAssembly/packages/boa/cpython.wasm?

rickyes:
> Is this from third-party https://github.com/rickyes/pipcook/tree/Compiling-CPython-to-WebAssembly/packages/boa/cpython.wasm?

Yes, cpython.wasm is a reference to [pyodide](https://github.com/iodide-project/pyodide), I'm looking for an easier way to build it.

yorkie:
> Yes, cpython.wasm is a reference to [pyodide](https://github.com/iodide-project/pyodide), I'm looking for an easier way to build it.

I think we must keep the name pyodide for third-party dependencies.

rickyes:
> > Yes, cpython.wasm is a reference to [pyodide](https://github.com/iodide-project/pyodide), I'm looking for an easier way to build it.
> 
> I think we must keep the name pyodide for third-party dependencies.



> pyodide

Okay, if it comes down to using pyodide as the way to build wasm, I'll keep the name as is.

rickyes:
Currently using `pyodide` to compile `cpython` to get the `.asm.wasm` and `.asm.js` file, we can open another PR to write a multi-platform interface to `.asm.js`

yorkie:
How about download pre-build pyodide binary-only?

rickyes:
> How about download pre-build pyodide binary-only?

Good idea, we can download pyodide binary packages only when we compile wasm.

PR #238, opened 2020-06-15
costa: fix git-install protocol missing

yorkie:


packages/costa/src/runtime.ts
```diff
@@ -187,7 +189,7 @@ export class CostaRuntime {
       await spawnAsync('npm', [ 'init', '-y' ], npmExecOpts);
       await spawnAsync('npm', [ 'install', boaSrcPath, '-E' ], npmExecOpts);
     }
-    await spawnAsync('npm', [ 'install', `${pluginAbsName}`, '-E', '--production', '--verbose' ], npmExecOpts);
+    await spawnAsync('npm', [ 'install', `${pluginAbsName}`, '-E', '--production' ], npmExecOpts);
```
rickycao-qy, line 192: What's the reason to remove verbose

yorkie: This is unneeded actually, I added this flag for debugging at #233.

rickycao-qy:
LGTM

Issue #33, opened 2020-03-03
PipApp: the application framework for machine learning


yorkie:
The vision of Pipcook is to take the JavaScript developers and engineers into the world of machine learning quickly and seamlessly, then we're responsible for creating easy enough APIs.

In the Pipcook stack, the `pipcook-app` is to be defined the ML application, which abstracts some duplicated stuffs and hides low-level algorithm implementation which requires a learning curve for every ML rookie.

## APIs

Every module represents a type of dataset, and basically we provide some different methods for developers.

### module `ml`

This module is to create machine learning functions, it provides the core abilities to represent your machine learning application in an intuitive way.

##### interface `ml.Function`

To hide the ML details as possible, Pipcook lets your declare your functions for machine learning purpose in a specific type `ml.Function`, you can create a `ml.Function` via the following `create()` function.

Internally, the Pipcook compiler parses the applications, then generates the training code via the `ml.Function` instances, and replaces these slots with model generated inferences.

##### interface `ml.FunctionImpl(arg: data.MLType)`

This interface is to describe the internal machine learning internals for applications, and it accepts an argument in `data.MLType` as the input, however the output's type is not required.

##### `create(fn: ml.FunctionImpl): ml.Function`

This is to create the above `ml.Function` with a `ml.FunctionImpl` object.

```ts
const mlfunc: ml.Function = ml.create((input: ml.ImageType) => {
  // call other ML Application APIs here and return
});

// ...
mlfunc(new ml.ImageType(...)); // call this function anywhere.
```

### module `data`

This module is to declare all types for your application's I/O.

##### interface `data.MLType`

It's the base interface to tell the Pipcook compiler a type for ML.

##### interface `data.ImageType` extends `data.MLType`

It represents the image type for given `ml.Function` I/O.

##### interface `data.TextType` extends `data.MLType`

It represents the text type for given `ml.Function` I/O.

### module `vision`

This module provides vision-related functions like image classification and object detection.

##### interface `vision.Position2D`

it represents the position in 2d for object detections:

- `label` {string} the label string represents the object's type.
- `left` {number} the left of detected object in pixel.
- `top` {number} the top of detected object in pixel.
- `height` {number} the height of detected object.
- `width` {number} the width of detected object.

##### `classify(img: ImageType): string`

It recognizes the type of image, and returns the type string.

```ts
ml.create((img: data.ImageType) => {
  const label = vision.classify(img); // returns the label
});
```

[...]

PR #241, opened 2020-06-16
add PipApp implementation

yorkie:
This implements #33 by adding a new package `@pipcook/app` which compiles the script source and outputs pipelines and models. And the current script version has the following restrictions:

- single-file only.
- top-level declaration for `learnable` blocks.

A quick start to show how PipApp looks like:

```ts
import { createLearnable, nlp, types } from '@pipcook/app';
const isCooking = createLearnable(async function isCookingImpl(sentence: string) {
  return (await nlp.classify(sentence));
});

(async () => {
  console.log(await isCooking('test'));
})();
```

The pull request TODO:

- [x] generates the pipeline configs.
- [x] creates pipelines and save to the current directory.
- [x] runs configured pipelines.
- [x] builds the executable and complete program.
- [x] documentation


yorkie:
Complete the saving pipelines.

packages/app/src/apis/executable.ts
```diff
@@ -0,0 +1,28 @@
...
+
+if (pathExistsSync(APP_MANIFEST)) {
+  manifest = readJSONSync(APP_MANIFEST);
+  if (manifest.executable === true) {
+    executable = true;
+  }
+}
+
+function dynamicModelExports(module: string, exports: any) {
+  manifest?.pipelines.filter((pipeline: any) => {
```
rickycao-qy, line 17: Is it possible to remove any type ~~

packages/app/src/compiler/pipelines/vision-classify.base.json
```diff
@@ -0,0 +1,35 @@
+{
+  "plugins": {
+    "dataCollect": {
+      "package": "./packages/plugins/data-collect/image-classification-data-collect",
+      "params": {
+        "url": "<TODO>"
```
rickycao-qy, line 6: nlp-classify.baes.json has the message 'input your CSV zip' and here the message is 'TODO'. This seems not consistent

packages/costa/src/runtime.ts
```diff
@@ -14,10 +14,15 @@ import {
   CondaConfig
 } from './index';
 
-import request from 'request-promise';
+import { get } from 'request-promise';
 import Debug from 'debug';
 
 const debug = Debug('costa.runtime');
+const UserAgent = [
```
rickycao-qy, line 21: Why do we want to include user-agent to identify browser here

yorkie: We can remove it now :)

yorkie:
@rickycao-qy resolved all nits :)

rickycao-qy:
LGTM

PR #244, opened 2020-06-16
fix: predict import plugin path with wrong path

Joker-Jelly:
Import plugin path need to point to path-to-pipcode-project/node_modules, not cwd


yorkie:
Thank you @Joker-Jelly :)

PR #245, opened 2020-06-16
record metadata of dataset 

rickycao-qy:
Metadata of dataset contains useful information about dataset itself.


packages/daemon/src/service/pipeline.ts
```diff
@@ -250,6 +250,10 @@ export class PipelineService {
       job.evaluatePass = result.pass;
       job.endTime = Date.now();
       job.status = PipelineStatus.SUCCESS;
+      const datasetVal = await runnable.valueOf(dataset) as UniDataset;
+      if (datasetVal?.metadata) {
+        job.dataDescription = JSON.stringify(datasetVal.metadata);
```
yorkie, line 255: How about just `dataset`?

rickycao-qy: sounds good

rickycao-qy: have already modified

PR #250, opened 2020-06-18
boa: fix the install script python

yorkie:
In this PR, we use the specific Python binary for interrupting the Pip and Conda script to avoid environment issues.

### DIFF
--- a/packages/boa/lib/index.js
+++ b/packages/boa/lib/index.js
@@ -14,7 +14,7 @@ const IterIdxForSeqSymbol = Symbol('The iteration index for sequence');
 // eslint-disable-next-line no-sync
 const condaPath = fs.readFileSync(path.join(__dirname, '../.CONDA_INSTALL_DIR'), 'utf8');
 // eslint-disable-next-line no-process-env
-process.env.PYTHONHOME = [ process.env.PYTHONHOME || '', condaPath ].join(':');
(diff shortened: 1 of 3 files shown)


packages/boa/lib/index.js
```diff
@@ -10,6 +10,15 @@ const DelegatorLoader = require('./delegator-loader');
 // internal symbols
 const IterIdxForSeqSymbol = Symbol('The iteration index for sequence');
 
+// read the conda path from the .CONDA_INSTALL_DIR
+// eslint-disable-next-line no-sync
+const condaPath = fs.readFileSync(path.join(__dirname, '../.CONDA_INSTALL_DIR'), 'utf8');
```
FeelyChau, line 15:
```suggestion
const condaPath = fs.readFileSync(path.join(__dirname, '..', '.CONDA_INSTALL_DIR'), 'utf8');
```
for more cross-platform adaptability.

FeelyChau: And could we use async function？

yorkie: unfortunately, we have to use this function here, because the top-level await is not supported.

yorkie: > adaptability

The `path.join` resolved the cross-platform issue in this way, you could try:

```
> path.win32.join('foobar', '123/test')
'foobar\\123\\test'
```

PR #254, opened 2020-06-21
plugin: modified data-process-plugin

nerffei:
1. add standardize method
2. change original implemention and config file

### DIFF
--- a/packages/core/package-lock.json
+++ b/packages/core/package-lock.json
@@ -4,6 +4,91 @@
 	"lockfileVersion": 1,
 	"requires": true,
 	"dependencies": {
+		"@tensorflow/tfjs": {
+			"version": "2.0.1",
+			"resolved": "https://registry.npmjs.org/@tensorflow/tfjs/-/tfjs-2.0.1.tgz",
+			"integrity": "sha512-XMI8gH1k0ushRR7hWAUXoy0WiAil2R7zOU6oz9AMioXTFK2cEn08xgpTXS7KLZUS4CnXN4NHOKbYn3/Dp+CSTA==",
+			"requires": {
+				"@tensorflow/tfjs-backend-cpu": "2.0.1",
+				"@tensorflow/tfjs-backend-webgl": "2.0.1",
+				"@tensorflow/tfjs-converter": "2.0.1",
+				"@tensorflow/tfjs-core": "2.0.1",
+				"@tensorflow/tfjs-data": "2.0.1",
+				"@tensorflow/tfjs-layers": "2.0.1",
+				"core-js": "3",
+				"regenerator-runtime": "^0.13.5"
+			}
+		},
+		"@tensorflow/tfjs-backend-cpu": {
+			"version": "2.0.1",
+			"resolved": "https://registry.npmjs.org/@tensorflow/tfjs-backend-cpu/-/tfjs-backend-cpu-2.0.1.tgz",
+			"integrity": "sha512-ZTDdq+O6AgeOrkek42gmPWz2T0r8Y6dBGjEFWkCMLI/5v3KnkodUkHRQOUoIN5hiaPXnBp6425DpwT9CfxxJOg==",
+			"requires": {
+				"@types/seedrandom": "2.4.27",
+				"seedrandom": "2.4.3"
+			}
+		},
+		"@tensorflow/tfjs-backend-webgl": {
+			"version": "2.0.1",
+			"resolved": "https://registry.npmjs.org/@tensorflow/tfjs-backend-webgl/-/tfjs-backend-webgl-2.0.1.tgz",
+			"integrity": "sha512-Z0MNsbHyRBr2lHu4FKtmBgGGfmBOqB0zwN69d2uSEEvy/IsgtrGAwBVbB/Vv3hEK35RmgsaknKGergNugm23FQ==",
+			"requires": {
+				"@tensorflow/tfjs-backend-cpu": "2.0.1",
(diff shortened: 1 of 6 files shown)


rickycao-qy:
What's particular reason to use tfjs-node to process image instead of some image-process libraries. And meanwhile, i'm a little worried since pipcook-core will be installed for every plugin and tfjs-node installation will be very slowed cuz it's on aws.

yorkie:
I think it's good to move those functions to core only :)

nerffei:
> Hi Nerffei, thanks for the contribution. What's particular reason to use tfjs-node to process image instead of some image-process libraries. And meanwhile, i'm a little worried since pipcook-core will be installed for every plugin and tfjs-node installation will be very slowed cuz it's on aws. 

I have some reasons here

1.  Currently data-process-plugin is processing pictures one by one. We can read images as  tensors, so that processing on batch 
2. When the image processing is completed，data process plugin is saved the image currently. If using tensor, we can keep it in memory for next train dierctly, or the other saved format.
3. I didn't find standardize method in jimp.Tfjs-node process image is hard. You are right, Maybe I should use both.

Indeed, Tfjs-node's installation is slow

I want to change the current data-process-plugin. Add more methods and  process on batch. 

I don't know how to do. this is a little try. I need suggestions

yorkie:
> When the image processing is completed，data process plugin is saved the image currently. If using tensor, we can keep it in memory for next train dierctly, or the other saved format.

If we keep all images in tensor, we also need to care about memory usage?

yorkie:
> Currently data-process-plugin is processing pictures one by one. We can read images as tensors, so that processing on batch.

@nerffei The input of data process plugin should keep the image type, could we use tensor in a plugin itself?

nerffei:
> > When the image processing is completed，data process plugin is saved the image currently. If using tensor, we can keep it in memory for next train dierctly, or the other saved format.
> 
> If we keep all images in tensor, we also need to care about memory usage?

目前, **data-access-plugin**: make dataloader -> done, **data-process-plugin**: read image -> process image -> save image -> done.然后,  **train-model-plugin**: read image again ->  make dataloader to "dataset" -> train -> done;

下面是个简单的流程，描述pytorch(其他框架不太熟悉) 的训练，估计大同小异吧
`
for batch in data_loader:
    const { images,labels }  = batch
    train(images, labels)
`
上面这个流程中，如果要对数据集的数据处理的话，是在data_loader yield batch的时候做处理

目前在pipcook中，是得到数据后，通过access后，再在data-process做这个事情，并且保存下来，然后放到模型中做训练。

这里核心不一样的地方在于：

一个是在data-loader拿数据，生成batch的时候，实时做处理。一个是在模型训练前，就把数据处理好。然后去训练

我观念更倾向于应该做前者。大致理由如下：

1. 有很多的随机处理，比如随机旋转，随机剪裁，比较希望每次拿batch的时候是不一样的，有一定增强样本的作用

2. 图片处理操作是很复杂的，分类任务可能不需要处理label。 以resize为例， 光resize图片是不够的，比如检测任务的话还要resize bbox， 分割任务要处理mask. 我觉得框架并不需要处理这种事情，用户处理自己的数据再来训练。

如果data-process是作用于data-access生成的dataloader之中。并且是对batch处理，不光能减少遍历图片一次，还能增加运算速度，还可以解决random操作(旋转，剪裁)等问题。如果这样的话， 内存完全取决于batch大小。

不过深度学习都吃内存。我自己在工作中，由于我武力霸占公司机子😄如果数据集不大的话，我都是把整个数据集放到内存里面的。

nerffei:
standardize 是很常见的需求，是需要加的，tfjs node确实安装慢，还出错，好像CI的错误都是来自这个。不过jimp没这个方法，你们有什么建议么

yorkie:
I think this problem should be looked at in different ways. On the one hand, it is the form of data stream processing, on the other hand, whether to use memory to reduce I/O operations.

1. I agree that when we get the data, need to pass it to the next process ASAP, instead of waiting for all the data to be processed before proceeding to the next step, but this will also have higher requirements for the design of the pipeline and The training framework itself also needs to support this training method, but on the whole, making each Pipeline train faster is the goal that Pipcook will work hard for.

2. Using our plugin to use Node.js/Web Streams may solve this problem and put the data stream in memory, but it is not necessary to put all the data in memory.

> 我觉得这个问题要分方面看，一方面是数据流的处理形式，一方面是是否使用内存来减少 I/O 操作，我的观点如下：
>
> 1. 我非常同意拿到数据后，尽快得丢到后面的流程，而不是等待所有的数据处理完再进行下一步，但这也会对 Pipeline 的设计有较高的要求，以及对训练框架本身也需要支持这种训练方式，不过整体来说使每条 Pipeline 训练得更快就是 Pipcook 会去努力的目标。
> 2. 将我们插件使用 Node.js/Web Streams 也许可以解决这个问题，将数据流放到内存里，但是不需要将数据都放到内存。

yorkie:
> standardize 是很常见的需求，是需要加的，tfjs node确实安装慢，还出错，好像CI的错误都是来自这个。不过jimp没这个方法，我去找找别的实现，你们有什么建议么

@rickycao-qy How about using boa to use a similar implementation in Python?

nerffei:
> > standardize 是很常见的需求，是需要加的，tfjs node确实安装慢，还出错，好像CI的错误都是来自这个。不过jimp没这个方法，我去找找别的实现，你们有什么建议么
> 
> @rickycao-qy How about using boa to use a similar implementation in Python?

我在jimp上扩展一下试试。。？

PR #255, opened 2020-06-22
add shuffle to dataset

rickycao-qy:
The data in dataset should be shuffled to achieve better performance


packages/core/src/utils/public.ts
```diff
@@ -290,3 +290,15 @@ export function getOsInfo() {
     });
   });
 }
+
+/**
+ * shuffle array
+ * @param array
```
yorkie, line 296: Could we have more documentation for this method?

packages/core/src/utils/public.ts
```diff
@@ -290,3 +290,15 @@ export function getOsInfo() {
...
+ */
+export function shuffle(array: any[]) {
```
yorkie, line 298: Need a return value and unit tests for this function :)

packages/core/src/utils/public.ts
rickycao-qy: done

packages/core/src/utils/public.ts
rickycao-qy: shuffle will just operate original array (just like python's built-in shuffle function). I have added unit tests

packages/core/src/utils/public.ts
```diff
@@ -290,3 +290,15 @@ export function getOsInfo() {
...
+/**
+ * Shuffles array in place. ES6 version. This method is based on Fisher-Yates shuffle algorithm
+ * @param array An array containing the items.
+ */
+export function shuffle(array: any[]) {
+  for (let i = array.length - 1; i > 0; i--) {
+    const j = Math.floor(Math.random() * (i + 1));
+    [ array[i], array[j] ] = [ array[j], array[i] ];
+  }
+  return array;
```
yorkie, line 303: @rickycao-qy as you commented at https://github.com/alibaba/pipcook/pull/255#discussion_r443347354, we need to remove this return-statement to keep consistent.

rickycao-qy: done

PR #256, opened 2020-06-22
model: add text creation model based on tfjs

yorkie:


yorkie:
@rickycao-qy @FeelyChau This might be ready, may I ask you to have reviews?

rickycao-qy:
LGTM

PR #269, opened 2020-06-24
cli, daemon: add custom registry support

yorkie:
Added a new field `npmRegistryPrefix ` at `~/.pipcook/daemon.config.json`, it's used to configure the npm registry prefix for installing plugins via NPM.


packages/daemon/src/boot/plugin.ts
```diff
@@ -14,10 +16,19 @@ export default class PluginRuntime {
 
   @init()
   async connect(): Promise<void> {
+    let npmRegistryPrefix = 'https://registry.npmjs.com/';
+    try {
+      const config = await readJSON(PIPCOOK_DAEMON_CONFIG);
+      npmRegistryPrefix = config.npmRegistryPrefix || npmRegistryPrefix;
+    } catch (err) {
+      console.warn(`${PIPCOOK_DAEMON_CONFIG} not existed.`);
```
FeelyChau, line 24:
```suggestion
      console.warn(`read ${PIPCOOK_DAEMON_CONFIG} error: ${err.message}.`);
```

yorkie:
> And, could extended this configuration to takes effect on pipcook init?

The file is called `daemon.config.json`, so I think they should not be for the init subcommand.

> There is no way to set the custom registry unless edit the configuration file, maybe we cloud provide a command to configure this.

This is an internal config, just like we don't tell the user editing the env to set the tuna, this PR is to add the ability to change the registry to download the plugin, not exposing this ability to end-users.

rickycao-qy:
There are conflicts, please fix firstly.

PR #270, opened 2020-06-24
cli: run `npm init` for daemon/pipboard to avoid npm downgrading

yorkie:
This will run `npm init` before installing daemon and board, it will generate the package.json to avoid the npm node_modules downgrading, such as if there is a node_modules at `~/node_modules`, so all the pipcook packages will be installed at `~/node_modules` instead of `~/.pipcook`.

### DIFF
--- a/packages/cli/src/actions/init.ts
+++ b/packages/cli/src/actions/init.ts
@@ -15,7 +15,7 @@ const {
 } = Constants;
 
 async function npmInstall(npmClient: string, name: string, beta: boolean, cwd: string, env: NodeJS.ProcessEnv): Promise<void> {
-  exec(`${npmClient} init -f`, { cwd, env, stdio: 'ignore' });
+  exec(`"${npmClient}" init -f`, { cwd, env, stdio: 'ignore' });
   if (isLocal) {
     // FIXME(yorkie): manually make symlink to local development
     // This is because npm-install from path will not be compatible with lerna's node_modules/.staging dir.
--- a/packages/cli/src/actions/init.ts
+++ b/packages/cli/src/actions/init.ts
@@ -1,13 +1,13 @@
 import { join, dirname } from 'path';
 import { execSync as exec } from 'child_process';
-import fse, { symlink, readJson, ensureDir } from 'fs-extra';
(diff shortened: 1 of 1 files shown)


packages/cli/src/actions/init.ts
```diff
@@ -1,21 +1,24 @@
...
-import { isLocal, optionalNpmClients, daemonPackage, boardPackage } from '../config';
+import { optionalNpmClients, daemonPackage, boardPackage, isLocal } from '../config';
 
 const {
   BOA_CONDA_INDEX,
   BOA_CONDA_MIRROR
 } = Constants;
 
 async function npmInstall(npmClient: string, name: string, beta: boolean, cwd: string, env: NodeJS.ProcessEnv): Promise<void> {
+  exec(`${npmClient} init -f`, { cwd, env, stdio: 'ignore' });
```
FeelyChau, line 18:
```suggestion
  exec(`"${npmClient}" init -f`, { cwd, env, stdio: 'ignore' });
```

FeelyChau: For robustness, the first argument is `The command to run, with space-separated arguments`, if the `npmClient` has a space, the command failed.

packages/cli/src/actions/init.ts
```diff
@@ -24,8 +27,8 @@ async function npmInstall(npmClient: string, name: string, beta: boolean, cwd: s
     if (beta) {
       name = `${name}@beta`;
     }
-    const cmd = `${npmClient} install ${name} --force`;
-    console.info(`exec <${cmd}>`);
+    const cmd = `${npmClient} install ${name} -E --production`;
```
FeelyChau, line 30:
```suggestion
    const cmd = `"${npmClient}" install ${name} -E --production`;
```

packages/cli/src/actions/init.ts
FeelyChau: Look at:
https://nodejs.org/api/child_process.html#child_process_child_process_exec_command_options_callback

yorkie:
I will rebase this branch after #269 gets merged.

PR #272, opened 2020-06-24
cli: optimize the pipcook-daemon error

yorkie:
This optimizes the logs when the daemon starts failed:

- checking if the daemon is running via pidfile.
- outputs the access log when it failed.
- handles the daemon `exit`.

### DIFF
--- a/packages/cli/src/bin/pipcook-daemon.ts
+++ b/packages/cli/src/bin/pipcook-daemon.ts
@@ -23,7 +23,7 @@ async function start(): Promise<void> {
 
   // check if the process is running...
   if (await pathExists(DAEMON_PIDFILE)) {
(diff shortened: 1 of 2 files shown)


packages/cli/src/bin/pipcook-daemon.ts
```diff
@@ -20,6 +20,12 @@ interface DaemonBootstrapMessage {
 
 async function start(): Promise<void> {
   const spinner = ora();
+
+  // check if the process is running...
+  if (await pathExists(DAEMON_PIDFILE)) {
+    spinner.fail('daemon is running...');
```
FeelyChau, line 26:
```suggestion
    spinner.fail('starting daemon but ${DAEMON_PIDFILE} exists.');
```

PR #280, opened 2020-06-28
boa: add BOA_TUNA to simplify the mirror usage

yorkie:


packages/boa/tools/bip.js
```diff
@@ -1,10 +1,12 @@
 #!/usr/bin/env node
+
 const fs = require('fs');
 const { run, getCondaPath } = require('./utils');
 
 const args = process.argv.slice(2).join(' ');
 const condaPath = getCondaPath();
-if (fs.existsSync(condaPath)) {
-  run(`${condaPath}/bin/pip`, args);
+if (!fs.existsSync) {
```
rickycao-qy, line 8: why to check fs.existsSync function

yorkie: Sorry, I will fix this :p

PR #284, opened 2020-06-28
core: cleanup the type definitions

yorkie:


yorkie:
Removed the unnecessary type `TfJsLayersModel` because it equals to `UniModel`.

rickycao-qy:
LGTM

PR #285, opened 2020-06-28
need to reshape image to [1, x,x,x] before predict

rickycao-qy:
1. This bug is introduced by feeding images directly to model.predict in TensorFlow. Actually, we need to reshape image to [1,x,x,x].

2. We store strings inside database. So need to JSON.parse them before using.


packages/plugins/model-define/tfjs-simplecnn-model-define/src/index.ts
```diff
@@ -47,8 +47,8 @@ const simpleCnnModelDefine: ModelDefineType = async (data: ImageDataset, args: M
     outputShape = Object.keys(data.metadata.labelMap).length;
     labelMap = data.metadata.labelMap;
   } else {
-    const log = JSON.parse(fs.readFileSync(path.join(recoverPath, 'metadata.json'), 'utf8'));
-    labelMap = log.output.dataset.labelMap;
+    const log = JSON.parse(fs.readFileSync(path.join(recoverPath, '..', 'metadata.json'), 'utf8'));
```
yorkie, line 50: I'm not sure if we should read the metadata.json in a plugin, how about having some methods to `metadata.json` in "@pipcook/pipcook-core"? So that plugin maintainer will not care about how the metadata.json is.

Never mind, this change LGTM in this PR.

packages/plugins/model-define/tensorflow-mobilenet-model-define/src/index.ts
```diff
@@ -115,7 +115,8 @@ const mobilenetDefine: ModelDefineType = async (data: ImageDataset, args: ModelD
       image = tf.image.decode_jpeg(image, boa.kwargs({
         channels: 3
       }));
-      return this.model.predict(image).toString();
+      const shape = tf.shape(image).numpy();
+      return this.model.predict(tf.reshape(image, [1, shape[0], shape[1], shape[2]])).toString();
```
yorkie, line 119:
```js
[1].concat(shape.slice(0, 2))
```

rickycao-qy: done

PR #291, opened 2020-06-30
core: add `downloadAndExtractTo` utility method to simplify data collect plugin

yorkie:


packages/core/src/utils/public.ts
```diff
@@ -105,6 +104,36 @@ export function download(url: string, fileName: string) {
   });
 }
 
+/**
+ * Download the dataset from specific URL and extract to a generated path as the returned value.
+ * @param resUrl the resource url, support http://, https://, file:///.
+ */
+export async function downloadAndExtractTo(resUrl: string): Promise<string> {
```
FeelyChau, line 111: api name `downloadZipAndExtractTo` may be more explicit.

yorkie: See my test case: https://github.com/alibaba/pipcook/pull/291/files#diff-d99c36ca252dde5a7002823e4ded419cR83, this function supports JSON file :)

FeelyChau: This api name looks strange, if the api supports json file url, how cloud we extract it.

PR #294, opened 2020-07-01
boa: clean tmp files for pybind11 after make

FeelyChau:
After `npm install` for boa, there is a download directory left, should be removed.

### DIFF
--- a/packages/boa/package.json
+++ b/packages/boa/package.json
@@ -4,7 +4,7 @@
   "description": "Use Python modules seamlessly in Node.js",
   "main": "lib/index.js",
   "scripts": {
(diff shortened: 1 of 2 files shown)


packages/boa/package.json
```diff
@@ -4,7 +4,7 @@
   "description": "Use Python modules seamlessly in Node.js",
   "main": "lib/index.js",
   "scripts": {
-    "clean": "sh tools/clean-python.sh",
+    "clean": "sh tools/clean-python.sh; make -C ./pybind11/ clean",
```
yorkie, line 7: Keep this consistent with other scripts:

```suggestion
    "clean": "sh tools/clean-python.sh && make -C ./pybind11/ clean",
```

PR #297, opened 2020-07-02
pipboard: use tfjs to running models for showcase

yorkie:
This PR is to running these showcase models, it:

- decreases the prediction time cost from 20s to 1s.
- removes the dependency `@tensorflow/tfjs` from Pipboard, so `pipcook init` won't download tfjs now.


packages/pipboard/public/index.html
```diff
@@ -6,6 +6,7 @@
   <meta http-equiv="x-ua-compatible" content="ie=edge,chrome=1">
   <meta name="viewport" content="width=device-width">
   <title>Pipcook Board</title>
+  <script src="https://cdn.jsdelivr.net/npm/@tensorflow/tfjs/dist/tf.min.js"> </script>
```
rickycao-qy, line 9: Shall we use npm way to install tfjs? just to be consistent to any other dependencies

yorkie: Added @tensorflow/tfjs into dependencies of pipboard.

rickycao-qy:
LGTM :)

Issue #170, opened 2020-05-06
boa: from Y import X could not be imported by destructing Y


yorkie:
The Python code:

```python
from PIL import Image, ImageDraw
```

And:

```js
const { Image, ImageDraw } = boa.import('PIL'); // not working
const Image = boa.import('PIL.Image'); // working
const ImageDraw = boa.import('PIL.ImageDraw'); // working
```

That's great if the destructing is working for sub modules.

yorkie:
See [`PyImport_ImportModuleLevelObject`](https://docs.python.org/3/c-api/import.html#c.PyImport_ImportModuleLevelObject)'s description:

> Import a module. This is best described by referring to the built-in Python function __import__(), as the standard __import__() function calls this function directly.
>
> The return value is a new reference to the imported module or top-level package, or NULL with an exception set on failure. Like for __import__(), the return value when a submodule of a package was requested is normally the top-level package, unless a non-empty fromlist was given.

This function will help us to address the submodule issue, but it requires that we're able to get the destructing names when calling the `boa.import()`, does anyone know if there is the kind of hook at ECMA262?

An alternative library-scoped way is:

```js
const { Image, ImageDraw } = boa.import('PIL', ['Image', 'ImageDraw']);
```

> This is ugly so that we won't do, just works for this case :(

yorkie:
A better alternative algorithm:

- mark the object as imported when it's returned by `boa.import(name)`.
- re-implement the [Python's from-clause hitting algorithm](https://docs.python.org/3/reference/simple_stmts.html#the-import-statement) in Proxy object's `get` hook:
  - check if the imported module has an attribute by that name
  - if not, attempt to import a submodule with that name and then check the imported module again for that attribute
  - if the attribute is not found, ImportError is raised.
  - otherwise, a reference to that value is stored in the local namespace, using the name in the as clause if it is present, otherwise using the attribute name

Its advantage is to completely support the following call:

```js
const { Image, ImageDraw } = boa.import('PIL');
```

However the disadvantage is obvious, we have to re-implement partial of the Python's import algorithm. The best way is still to request a proposal for adding a destructing assignment hook.

yorkie:
Closing because this could not be completed logically.

PR #300, opened 2020-07-02
boa: fix import { X } from Y

FeelyChau:
fix issue: https://github.com/alibaba/pipcook/issues/170
```js
const { Image } = boa.import('PIL'); // working
const Image = boa.import('PIL.Image'); // working
const PIL = boa.import('PIL'); // working
console.log(PIL.Image); //working
```
The difference between Python and BOA implementations is:
* python
The submodule `Image` is created when module `PIL` is creating.
* boa
The module `PIL` is created, then sub module `Image` is requested, finally the `Image` is created and put into `PIL`.

### DIFF
--- a/packages/boa/src/core/object.cc
+++ b/packages/boa/src/core/object.cc
@@ -295,7 +295,7 @@ Napi::Value PythonObject::GetItem(const CallbackInfo &info) {
       std::string keystr = std::string(info[0].As<String>());
       if (PyModule_Check(_self.ptr())) {
         pybind::dict dict = pybind::reinterpret_borrow<pybind::dict>(
-          PyModule_GetDict(_self.ptr()));
+            PyModule_GetDict(_self.ptr()));
         if (dict.contains(keystr.c_str())) {
           pybind::object submodule = dict[keystr.c_str()];
           if (submodule && PyModule_Check(submodule.ptr()))
--- a/packages/boa/src/core/object.cc
+++ b/packages/boa/src/core/object.cc
@@ -294,7 +294,8 @@ Napi::Value PythonObject::GetItem(const CallbackInfo &info) {
       // otherwise, assert the key must be a string.
       std::string keystr = std::string(info[0].As<String>());
       if (PyModule_Check(_self.ptr())) {
-        pybind::dict dict = pybind::reinterpret_borrow<pybind::dict>(PyModule_GetDict(_self.ptr()));
+        pybind::dict dict = pybind::reinterpret_borrow<pybind::dict>(
+          PyModule_GetDict(_self.ptr()));
         if (dict.contains(keystr.c_str())) {
           pybind::object submodule = dict[keystr.c_str()];
           if (submodule && PyModule_Check(submodule.ptr()))
--- a/packages/boa/src/core/object.cc
+++ b/packages/boa/src/core/object.cc
@@ -293,10 +293,20 @@ Napi::Value PythonObject::GetItem(const CallbackInfo &info) {
     } else {
       // otherwise, assert the key must be a string.
       std::string keystr = std::string(info[0].As<String>());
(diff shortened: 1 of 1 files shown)


yorkie:
Sorry, your implementation affects all the Python objects, which is incorrect and inefficient, we only add this module check for the objects which is returned from `boa.import`.

FeelyChau:
> Sorry, your implementation affects all the Python objects, which is incorrect and inefficient, we only add this module check for the objects which is returned from `boa.import`.

No, it has been checked by this line:
https://github.com/alibaba/pipcook/pull/300/files#diff-6f3ee49d7146f79cfafee47de76b43b4R296
only module object will enter this scope.

yorkie:
> https://github.com/alibaba/pipcook/pull/300/files#diff-6f3ee49d7146f79cfafee47de76b43b4R296

I'm also -1 on this check here because the meaning of `PyModule_Check` does not belong to the `getItem`, and I recommend that we split the objects at the `boa.import` in JavaScript.

FeelyChau:
> > https://github.com/alibaba/pipcook/pull/300/files#diff-6f3ee49d7146f79cfafee47de76b43b4R296
> 
> I'm also -1 on this check here because the meaning of `PyModule_Check` does not belong to the `getItem`, and I recommend that we split the objects at the `boa.import` in JavaScript.

I will perform a performance analysis on this.

Before this PR, the operator [] for module object is incorrect, it will throw an error `TypeError: 'module' object is not subscriptable`, but be caught, and we do nothing for this.

FeelyChau:
> > > https://github.com/alibaba/pipcook/pull/300/files#diff-6f3ee49d7146f79cfafee47de76b43b4R296
> > 
> > 
> > I'm also -1 on this check here because the meaning of `PyModule_Check` does not belong to the `getItem`, and I recommend that we split the objects at the `boa.import` in JavaScript.
> 
> I will perform a performance analysis on this.
> 
> Before this PR, the operator [] for module object is incorrect, it will throw an error `TypeError: 'module' object is not subscriptable`, but be caught, and we do nothing for this.

Well, we just caught this error and print to debug, it will hidden the error information.
https://github.com/alibaba/pipcook/blob/master/packages/boa/lib/index.js#L377

I read the spec for import, the submodule won't be loaded automatically unless explicitly declared in `__init__.py`, so this implementation is not  incorrect.
I'll take a look at it again.

yorkie:
> I read the spec for import, the submodule won't be loaded automatically unless explicitly declared in __init__.py, so this implementation is not incorrect.

Why does Boa need to take care about the `__init__.py`? Python's import function will handle this.

FeelyChau:
> > I read the spec for import, the submodule won't be loaded automatically unless explicitly declared in **init**.py, so this implementation is not incorrect.
> 
> Why does Boa need to take care about the `__init__.py`? Python's import function will handle this.

Yeah, this is what I mean, it's another case. The point is, the following code is incorrect in cpython:
```python
import PIL
print(PIL.Image) # throw Error
```
let's go back to the node:
```js
/* case 1*/
const { Image } = boa.import('PIL'); // should work

/* case 2 */
const PIL = boa.import('PIL');
console.log(PIL.Image); // should throw error or get undefined
```
* case 1:
node import requires 2 step:
1. construct the import module (`PIL`)
2. destruct module object `PIL`, get sumodule `Image`
* case 2:
we get `PIL.Image` by 2 step too:
1. construct the import module (`PIL`)
2. get submodule `Image` by getter of `PIL`

Deconstruction is actually done through getter, so we can't distinguish between the two cases, so we can't expect the same code to behave the same in node environment.

yorkie:
I think the key problem is to split the import-clause with the dot-clause, however it seems that not supported at Node.js/v8.

FeelyChau:
I think this PR should be closed for now.

PR #310, opened 2020-07-06
plugins: update the glob wildcard to support more flexible searching for dataset

lijiajunxs:
zip解压包的多级文件目录兼容

### DIFF
--- a/packages/plugins/data-collect/csv-data-collect/src/index.ts
+++ b/packages/plugins/data-collect/csv-data-collect/src/index.ts
@@ -32,7 +32,7 @@ const textClassDataCollect: DataCollectType = async (args: ArgsType): Promise<vo
   await fs.ensureDir(dataDir);
 
   const tempDir = await downloadAndExtractTo(url);
-  const csvPaths = await glob(path.join(tempDir, '**','+(train|validation|test)', '*.csv'));
+  const csvPaths = await glob(path.join(tempDir, '**', '+(train|validation|test)', '*.csv'));
   const trainData: any[] = [];
   const validationData: any[] = [];
   const testData: any[] = [];
--- a/packages/plugins/data-collect/csv-data-collect/src/index.ts
+++ b/packages/plugins/data-collect/csv-data-collect/src/index.ts
@@ -32,7 +32,7 @@ const textClassDataCollect: DataCollectType = async (args: ArgsType): Promise<vo
   await fs.ensureDir(dataDir);
 
   const tempDir = await downloadAndExtractTo(url);
-  const csvPaths = await glob(path.join(tempDir, '+(train|validation|test)', '*.csv'));
+  const csvPaths = await glob(path.join(tempDir, '**','+(train|validation|test)', '*.csv'));
   const trainData: any[] = [];
   const validationData: any[] = [];
   const testData: any[] = [];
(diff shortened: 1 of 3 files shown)


packages/plugins/data-collect/csv-data-collect/src/index.ts
```diff
@@ -32,7 +32,7 @@ const textClassDataCollect: DataCollectType = async (args: ArgsType): Promise<vo
   await fs.ensureDir(dataDir);
 
   const tempDir = await downloadAndExtractTo(url);
-  const csvPaths = await glob(path.join(tempDir, '+(train|validation|test)', '*.csv'));
+  const csvPaths = await glob(path.join(tempDir, '**','+(train|validation|test)', '*.csv'));
```
yorkie, line 35:
```suggestion
  const csvPaths = await glob(path.join(tempDir, '**', '+(train|validation|test)', '*.csv'));
```

packages/plugins/data-collect/image-classification-data-collect/src/index.ts
```diff
@@ -58,7 +58,7 @@ const imageClassDataCollect: DataCollectType = async (args: ArgsType): Promise<v
   const imageDir = path.join(dataDir, 'images');
   console.log('unzip and collecting data...');
   await unZipData(url, imageDir);
-  const imagePaths = await glob(path.join(imageDir, '+(train|validation|test)', '*', '*.+(jpg|jpeg|png)'));
+  const imagePaths = await glob(path.join(imageDir, '**','+(train|validation|test)', '*', '*.+(jpg|jpeg|png)'));
```
yorkie, line 61:
```suggestion
  const imagePaths = await glob(path.join(imageDir, '**', '+(train|validation|test)', '*', '*.+(jpg|jpeg|png)'));
```

packages/plugins/data-collect/object-detection-coco-data-collect/src/index.ts
```diff
@@ -90,7 +90,7 @@ const imageDetectionDataCollect: DataCollectType = async (args: ArgsType): Promi
   const imageDir = path.join(dataDir, 'images');
   console.log('unzip and collecting data...');
   await unZipData(url, imageDir);
-  const annotationPaths = await glob(path.join(imageDir, '+(train|validation|test)', 'annotation.json'));
+  const annotationPaths = await glob(path.join(imageDir, '**','+(train|validation|test)', 'annotation.json'));
```
yorkie, line 93:
```suggestion
  const annotationPaths = await glob(path.join(imageDir, '**', '+(train|validation|test)', 'annotation.json'));
```

rickycao-qy:
Plugins like object-detection-pascalvoc-data-collect also have this unzipping problem, though it does not use a wildcard. Is it a good idea to fix it in this PR?

yorkie:
> Plugins like object-detection-pascalvoc-data-collect also have this unzipping problem, though it does not use a wildcard. Is it a good idea to fix it in this PR?

@rickycao-qy It's a good first contribution, IMO we need to keep this kind of PR to be minimal so that the new contribution will be more friendly :)

yorkie:
The CI is failing on this test case https://github.com/alibaba/pipcook/blob/master/packages/core/src/utils/public_test.ts#L88-L95 due to the network issue again, maybe we need a patch to set a timeout.

/cc @FeelyChau

yorkie:
@rickycao-qy could you please have a double check on this PR?

PR #321, opened 2020-07-09
costa: support in-memory cache for npm registry requests

yorkie:
This optimizes the request to npm registries by using HTTP cache in memory.


packages/costa/src/runtime.ts
```diff
@@ -108,6 +108,58 @@ function createRequirements(name: string, config: CondaConfig): string[] {
...
+interface HTTPCache {
+  etag: string;
+  lastModified: string;
+  body: any;
+}
+
+/**
+ * `HTTPCache` with the uri key.
+ */
+const httpCacheByURI: Record<string, HTTPCache> = {};
```
FeelyChau, line 124: Is the cache size needs to be limited for preventing excessive memory usage?

yorkie: Good catch! I added an `LRUCache` for enabling this limit, please have a double review :)

packages/costa/src/lrucache.ts
```diff
@@ -0,0 +1,27 @@
+export default class LruCache<T> {
+  private values: Map<string, T> = new Map<string, T>();
```
FeelyChau, line 2:
```suggestion
  private values = new Map<string, T>();
```
The type of values can derivate by the right value.

PR #326, opened 2020-07-09
daemon: return immediately if job run without verbose

FeelyChau:
return immediately if job run without verbose.

### DIFF
--- a/packages/daemon/src/app/controller/job.ts
+++ b/packages/daemon/src/app/controller/job.ts
@@ -1,12 +1,12 @@
 import { Context, controller, inject, provide, get } from 'midway';
 import { successRes, failRes } from '../../utils/response';
-import { PipelineService, PluginInfo } from '../../service/pipeline';
+import { PipelineService } from '../../service/pipeline';
 import { parseConfig } from '../../runner/helper';
 import ServerSentEmitter from '../../utils/emitter';
 import { JobModel } from '../../model/job';
 import { PluginManager } from '../../service/plugin';
 import { createReadStream } from 'fs';
-import { PluginTypeI, PipelineStatus } from '@pipcook/pipcook-core';
+import { PipelineStatus } from '@pipcook/pipcook-core';
 
 @provide()
 @controller('/job')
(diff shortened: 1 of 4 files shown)


packages/daemon/src/app/controller/job.ts
```diff
@@ -60,15 +61,18 @@ export class JobController {
         sse.finish();
       }
     } else {
-      try {
-        const plugins = await this.pipelineService.installPlugins(job, cwd, pyIndex);
+      const run = async () => {
```
yorkie, line 64: Why do you create this function and it gets called once.

FeelyChau: Moved into `nextTick`

packages/daemon/src/app/controller/job.ts
```diff
@@ -60,19 +61,21 @@ export class JobController {
...
-        return failRes(this.ctx, {
-          message: err?.message,
-          data: job
-        });
-      }
       successRes(this.ctx, {
         message: 'create pipeline and jobs successfully',
         data: job
       }, 201);
+      process.nextTick(async () => {
```
yorkie, line 68: It's good to keep it the same position for minimum change.

packages/daemon/src/app/controller/job.ts
```diff
@@ -60,19 +61,21 @@ export class JobController {
...
+        let plugins: Partial<Record<PluginTypeI, PluginInfo>>;
+        try {
+          plugins = await this.pipelineService.installPlugins(job, cwd, pyIndex);
+        } catch (err) {
+          job.status = PipelineStatus.FAIL;
+          job.error = err.message;
+          return job.save();
+        }
+        this.pipelineService.startJob(job, cwd, plugins);
```
yorkie, line 77: Why do we move the `startJob()` outside of the try-clause?

FeelyChau: `startJob` do not throw any error.

packages/daemon/src/app/controller/job.ts
```diff
@@ -60,15 +61,17 @@ export class JobController {
         sse.finish();
       }
     } else {
-      try {
-        const plugins = await this.pipelineService.installPlugins(job, cwd, pyIndex);
+      process.nextTick(async () => {
+        let plugins: Partial<Record<PluginTypeI, PluginInfo>>;
```
yorkie, line 65: Ditto

FeelyChau:
The apis refactored in #380 , so we should close this PR.

PR #330, opened 2020-07-10
costa: add bootstrap script for bootstrap

yorkie:
To run the bootstrap benchmark for costa:

```sh
$ cd packages/costa
$ make -C benchmark
time node ./bootstrap.js
  costa.runnable make sure the component dir is existed. +0ms
  costa.runnable bootstrap a new process for c00b5cf0-c26f-11ea-93dc-2165bcacf22a. +4ms
  costa.runnable sent handshake for c00b5cf0-c26f-11ea-93dc-2165bcacf22a, and wait for response +9ms
  costa.runnable the runnable(c00b5cf0-c26f-11ea-93dc-2165bcacf22a) has been destroyed with(code=0, signal=null). +106ms
        0.44 real         0.40 user         0.08 sys
```

### DIFF
--- a/packages/costa/benchmark/makefile
+++ b/packages/costa/benchmark/makefile
@@ -1,0 +1,4 @@
+export DEBUG=costa*
+
+bootstrap:
+	time node ./bootstrap.js
--- a/packages/costa/benchmark/bootstrap.js
+++ b/packages/costa/benchmark/bootstrap.js
@@ -1,0 +1,18 @@
+'use strict';
+
+const { join } = require('path');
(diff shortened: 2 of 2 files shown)


packages/costa/benchmark/makefile
```diff
@@ -0,0 +1,4 @@
+export DEBUG=costa*
```
FeelyChau, line 1: Why not use the npm script but make?

yorkie: Makefile is more simple for this use-case, and this won't be running at userland.

PR #333, opened 2020-07-10
ci: speed up with cache

FeelyChau:


FeelyChau:
The speed of github action cache is not unstable, we need to reappraise.

Issue #337, opened 2020-07-13
cli: incorrect cli output when daemon is crashed


yorkie:
The CLI outputs the following when the daemon is crashed:

```sh
daemon is not started, run "pipcook daemon restart"
```

The correct behavior should be:

```sh
daemon is crashed with an error:
[stacktrace of daemon process]
```

yorkie:
This is fixed by #345.

PR #345, opened 2020-07-15
cli: correct the error message when receiving http errors.

yorkie:
To reproduce the bug that this PR tries to fix, just run:

```sh
$ pipcook run https://foobar
```

### DIFF
--- a/packages/cli/src/request.ts
+++ b/packages/cli/src/request.ts
@@ -5,6 +5,10 @@ import { ora } from './utils';
 
 export type RequestParams = Record<string, any>;
 export type ResponseParams = Record<string, any>;
+type ErrorEvent = {
+  status?: number;
+  message: string;
+} & Event;
 
 function createGeneralRequest(agent: Function): Function {
   const spinner = ora();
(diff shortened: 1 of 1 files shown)


yorkie:
And this fixes #337.

packages/cli/src/request.ts
```diff
@@ -51,12 +55,18 @@ export const listen = async (host: string, params?: RequestParams, handlers?: Re
...
         es.close();
         clearTimeout(timeoutHandle);
-        console.error('daemon is not started, run "pipcook daemon start"');
         es.removeEventListener('error', onerror);
+
+        // print error
+        if (typeof e?.status === 'number') {
+          console.error('occurrs an daemon error with the following response:\n', e);
+        } else {
+          console.error(`daemon is not started(${e.message}), run "pipcook daemon start".`);
```
FeelyChau, line 68: if `e` is undefined, throw an error here.

yorkie: See https://github.com/EventSource/eventsource/blob/master/lib/eventsource.js#L53, https://github.com/EventSource/eventsource/blob/master/lib/eventsource.js#L143 and https://github.com/EventSource/eventsource/blob/master/lib/eventsource.js#L162, the upstream library `eventsource` guarantee this won't be a undefined value.

FeelyChau: If `e` is not undefined, why we check it by `if (typeof e?.status === 'number') {`?

yorkie:
Updated, thank you @FeelyChau :)

PR #347, opened 2020-07-15
boa: skip the esm tests for node >= 14.5.0

yorkie:
See https://github.com/nodejs/node/pull/33501, At Node.js 14.5.0, it removes the dynamic module format, so we need to make compatible works for the new with esm loader.

Node.js Release PR: https://github.com/nodejs/node/pull/34093


yorkie:
Open an issue https://github.com/alibaba/pipcook/issues/348 for tracking the future support work.

PR #351, opened 2020-07-16
check if tty is enabled when spinner log

rickycao-qy:


yorkie:
@rickycao-qy May I ask you to rebase because @FeelyChau has some updates?

rickycao-qy:
> @rickycao-qy May I ask you to rebase because @FeelyChau has some updates?

done

packages/cli/src/bin/pipcook-daemon.ts
```diff
@@ -19,14 +19,12 @@ interface DaemonBootstrapMessage {
 }
 
 async function start(): Promise<void> {
-  const spinner = ora();
-
   // check if the process is running...
   if (await pathExists(DAEMON_PIDFILE)) {
-    spinner.fail(`starting daemon but ${DAEMON_PIDFILE} exists.`);
+    logFail(`starting daemon but ${DAEMON_PIDFILE} exists.`);
     return;
```
FeelyChau, line 25: If the command fail, we should exit with error code.

packages/cli/src/bin/pipcook-daemon.ts
```diff
@@ -37,31 +35,30 @@ async function start(): Promise<void> {
     if (message.event === 'ready') {
       daemon.disconnect();
       daemon.unref();
-      spinner.succeed(`Pipcook is on http://localhost:${message.data.listen}`);
+      logSuccess(`Pipcook is on http://localhost:${message.data.listen}`);
     }
   });
   daemon.on('exit', async (code: number) => {
-    spinner.fail(`Pipcook daemon starts failed with code(${code}).`);
+    logFail(`Pipcook daemon starts failed with code(${code}).`);
```
FeelyChau, line 42: Ditto.

packages/cli/src/bin/pipcook-job.ts
```diff
@@ -17,17 +17,15 @@ async function list(): Promise<void> {
...
+  logSuccess('remove jobs succeeded');
 }
 
 async function stop(id: string): Promise<void> {
-  const spinner = ora();
-  spinner.start('stopping jobs...');
+  logStart('stopping jobs...');
   const err = await get(`${route.job}/stop`, { id });
-  err ? spinner.fail(err.message) : spinner.succeed('stop job succeeded');
+  err ? logFail(err.message) : logSuccess('stop job succeeded');
```
FeelyChau, line 28: Ditto.

packages/cli/src/pipeline.ts
```diff
@@ -33,29 +32,29 @@ export async function install(filename: string, opts: any): Promise<void> {
...
+          logSuccess(`plugin (${name}@${version}) is installed`);
         },
         'error': (e: MessageEvent) => {
-          spinner.fail(`occurrs an error ${e.data}`);
+          logFail(`occurrs an error ${e.data}`);
           reject(new TypeError(e.data));
         },
         'finished': () => {
-          spinner.succeed('all plugins installed');
+          logFail('all plugins installed');
```
FeelyChau, line 57: Should be `logSucceed`?

packages/cli/src/actions/dev-plugin.ts
```diff
@@ -3,9 +3,8 @@ import chalk from 'chalk';
 import path from 'path';
 import { constants } from '@pipcook/pipcook-core';
 import { DevPluginCommandHandler } from '../types';
-import { ora } from '../utils';
+import { logFail } from '../utils';
```
yorkie, line 6: Shall we use `abort()` instead?

packages/cli/src/actions/dev-plugin.ts
```diff
@@ -27,7 +26,7 @@ const devPlugin: DevPluginCommandHandler = async ({ type, name }) => {
   try {
     dirname = path.join(process.cwd(), name);
     if (fse.existsSync(dirname)) {
-      spinner.fail(`a directory or file called ${name} already exists. Please use a new working directory`);
+      logFail(`a directory or file called ${name} already exists. Please use a new working directory`);
```
yorkie, line 29: As I commented before, we can use `abort` instead of the above 2 lines which calls `process.exit` internally.

packages/cli/src/actions/init.ts
```diff
@@ -5,7 +5,7 @@ import { prompt } from 'inquirer';
 import { sync } from 'command-exists';
 import { constants as CoreConstants } from '@pipcook/pipcook-core';
 
-import { ora, Constants } from '../utils';
+import { Constants, logInfo, logFail, logSuccess } from '../utils';
```
yorkie, line 8: I would suggest moving the log related functions to a separate object `logger` to simplify these names.

yorkie: And the function `logSuccess` has not yet used :)

packages/cli/src/utils.ts
```diff
@@ -70,16 +72,36 @@ export async function parseConfigFilename(filename: string): Promise<string> {
...
+export function logSuccess(message: string) {
+  spinner.succeed(message);
+}
+
+export function logFail(message: string) {
+  spinner.fail(message);
+}
+
+export function logStart(message: string) {
+  if (process.stdout.columns == 0 && process.stdout.rows == 0) {
```
yorkie, line 84: How about the following to define the logger methods:

```js
class Logger {
  success();
  fail();
  start();
  info();
  warn();
}
class DefaultLogger extends Logger {
  start() { console.log(...) }
}
class TtyLogger extends Logger {
  start() { this.spinner.start(...) }
}
// set the logger by the environment
let logger = new DefaultLogger();
const { columns, rows, isTTY } = process.stdout;
if (isTTY && columns > 0 && rows > 0) {
  logger = new TtyLogger();
}
export logger;
```

Wrapping the TTY mode in class might be more maintainable for extending other methods in the future :)

packages/cli/src/utils.ts
```diff
@@ -70,16 +72,36 @@ export async function parseConfigFilename(filename: string): Promise<string> {
...
 }
 
 /**
  * leave a message and abort the process
  * @param spinner spinner object
  * @param message last message
  */
-export function abort(spinner: realOra.Ora, message: string): void {
-  spinner.fail(message);
+export function abort(message: string): void {
```
yorkie, line 104: I think it's good to add an optional parameter `code` or `codestr` to specify which code will be passed to `process.exit`.

rickycao-qy:
1. Have already move util functions to util class Logger
2. merge abort() and logFail into one function
3. can specify the error code when call logFail

Please check again pls:) @yorkie @FeelyChau

packages/cli/src/utils.ts
```diff
@@ -70,16 +70,44 @@ export async function parseConfigFilename(filename: string): Promise<string> {
...
+
+  constructor() {
+    this.spinner = realOra({ stream: process.stdout });
+  }
+
+  success(message: string) {
+    this.spinner.succeed(message);
+  }
+
+  fail(message: string, code?: number) {
```
yorkie, line 84: How about add 1 as the default value?

packages/cli/src/utils.ts
```diff
@@ -70,16 +70,44 @@ export async function parseConfigFilename(filename: string): Promise<string> {
   return filename;
 }
 
-export function ora(opts?: realOra.Options): realOra.Ora {
-  return realOra({ stream: process.stdout, ...opts });
+class Logger {
+  spinner: realOra.Ora;
```
yorkie, line 74: Shall we have the spinner at `TtyLogger`?

rickycao-qy: info, fail, success method still work in non-tty environment. I think spinner should exists in super class

packages/cli/src/utils.ts
rickycao-qy: There are some cases that do not need to exit the process immediately. If we specify default value, then we need to give parameter with 'undefined' value for some cases which seem not good

yorkie: Maintaining both `abort(msg, code?)` and `fail(msg)` might be more reasonable to distinguish these use-cases.

packages/cli/src/utils.ts
yorkie: IMHO, using `ora` means more interactive in a terminal, otherwise, we need the plain stdout/stderr instead.

packages/cli/src/utils.ts
rickycao-qy: Isn't is quite weird to have something called 'abort' in logger?

yorkie: `abort()` should be in utils instead of logger.

packages/cli/src/pipeline.ts
```diff
@@ -64,15 +62,13 @@ export async function install(filename: string, opts: any): Promise<void> {
...
   const cwd = process.cwd();
 
-  spinner.start('start running the pipeline...');
+  logger.start('start running the pipeline...');
   try {
     filename = await parseConfigFilename(filename);
   } catch (err) {
-    spinner.fail(err.message);
-    return process.exit(1);
+    return logger.fail(err.message);
```
FeelyChau, line 71: Missing the error code.

packages/cli/src/utils.ts
rickycao-qy: still does not make sense to use spinner in abort function and move abort function outside of logger. 'fail' already means that the program should exit.

packages/cli/src/utils.ts
rickycao-qy: done

packages/cli/src/pipeline.ts
rickycao-qy: done

packages/cli/src/utils.ts
yorkie: Ok, shall we add a new parameter to tell the function not exiting immediately? The current implementation looks implicit a bit.

rickycao-qy: done

PR #355, opened 2020-07-18
lint: add checking for unused vars and unused args

FeelyChau:
Add the rules for unused vars and args.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -22,6 +22,7 @@ module.exports = {
         'functions': false,
       }
     ],
+    "@typescript-eslint/no-unused-vars": [ "error", { "argsIgnorePattern": "^_", "varsIgnorePattern": "^_" } ],
     'comma-spacing': [
       'error',
       {
@@ -46,6 +47,6 @@ module.exports = {
     'block-spacing': 'error',
     'brace-style': [ 'error', '1tbs', { allowSingleLine: true } ],
     'comma-style': 'error',
-    "@typescript-eslint/no-unused-vars": [ "error", { "argsIgnorePattern": "^_", "varsIgnorePattern": "^_" } ]
+    "no-unused-vars": [ "error", { "argsIgnorePattern": "^_", "varsIgnorePattern": "^_" } ],
   }
 };
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -45,6 +45,7 @@ module.exports = {
     'arrow-spacing': [ 'error', { before: true, after: true } ],
     'block-spacing': 'error',
     'brace-style': [ 'error', '1tbs', { allowSingleLine: true } ],
-    'comma-style': 'error'
+    'comma-style': 'error',
+    "@typescript-eslint/no-unused-vars": [ "error", { "argsIgnorePattern": "^_", "varsIgnorePattern": "^_" } ]
   }
 };
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -47,6 +47,6 @@ module.exports = {
     'block-spacing': 'error',
     'brace-style': [ 'error', '1tbs', { allowSingleLine: true } ],
     'comma-style': 'error',
-    "no-unused-vars": [ "off" ], // Use @typescript-eslint/no-unused-vars instead, otherwise there will be checked twice
+    'no-unused-vars': [ 'off' ], // Use @typescript-eslint/no-unused-vars instead, otherwise there will be checked twice
   }
 };
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -22,7 +22,7 @@ module.exports = {
         'functions': false,
       }
     ],
-    "@typescript-eslint/no-unused-vars": [ "error", { "argsIgnorePattern": "^_", "varsIgnorePattern": "^_" } ],
+    "@typescript-eslint/no-unused-vars": [ "error" ],
     'comma-spacing': [
       'error',
       {
@@ -47,6 +47,6 @@ module.exports = {
     'block-spacing': 'error',
     'brace-style': [ 'error', '1tbs', { allowSingleLine: true } ],
     'comma-style': 'error',
-    "no-unused-vars": [ "error", { "argsIgnorePattern": "^_", "varsIgnorePattern": "^_" } ],
+    "no-unused-vars": [ "off" ], // Use @typescript-eslint/no-unused-vars instead, otherwise there will be checked twice
   }
 };
--- a/packages/app/src/apis/nlp.ts
+++ b/packages/app/src/apis/nlp.ts
@@ -1,43 +1,45 @@
+/*eslint no-unused-vars: "off"*/
+/*eslint @typescript-eslint/no-unused-vars: "off"*/
 import { NamedEntityProp, TextAnswer, TranslationResult } from './types';
 import { dynamicModelExports } from './executable';
 
 /**
  * Classify the given text input, it returns the label of this input.
- * @param _input the input text.
+ * @param input the input text.
  */
-export async function classify(_input: string): Promise<string> {
+export async function classify(input: string): Promise<string> {
   throw new TypeError('not trained method');
 }
 
 /**
  * Get the named entity by a sentence.
- * @param _sentence the sentence for named entity
+ * @param sentence the sentence for named entity
  */
-export async function getNamedEntity(_sentence: string): Promise<NamedEntityProp> {
+export async function getNamedEntity(sentence: string): Promise<NamedEntityProp> {
   throw new TypeError('not trained method');
 }
 
 /**
  * Get answers by given question, every anwser has the text and its confidence.
- * @param _question the question text.
+ * @param question the question text.
  */
-export async function getAnswers(_question: string): Promise<TextAnswer[]> {
+export async function getAnswers(question: string): Promise<TextAnswer[]> {
   throw new TypeError('not trained method');
 }
 
 /**
  * Get the summary of the given input.
- * @param _input the input text.
+ * @param input the input text.
  */
-export async function getSummary(_input: string): Promise<string> {
+export async function getSummary(input: string): Promise<string> {
   throw new TypeError('not trained method');
 }
 
 /**
  * Returns the translated language.
- * @param _input the input text to be translated.
+ * @param input the input text to be translated.
  */
-export async function translate(_input: string): Promise<TranslationResult> {
+export async function translate(input: string): Promise<TranslationResult> {
   throw new TypeError('not trained method');
 }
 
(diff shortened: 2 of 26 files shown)


.eslintrc.js
```diff
@@ -45,6 +45,7 @@ module.exports = {
     'arrow-spacing': [ 'error', { before: true, after: true } ],
     'block-spacing': 'error',
     'brace-style': [ 'error', '1tbs', { allowSingleLine: true } ],
-    'comma-style': 'error'
+    'comma-style': 'error',
+    "@typescript-eslint/no-unused-vars": [ "error", { "argsIgnorePattern": "^_", "varsIgnorePattern": "^_" } ]
```
yorkie, line 49: Sort in the alphabet, this line should be put https://github.com/alibaba/pipcook/blob/d1923a140fe84f6e773c4947b0a8489bef030a86/.eslintrc.js#L24.

packages/app/src/apis/nlp.ts
```diff
@@ -3,41 +3,41 @@ import { dynamicModelExports } from './executable';
 
 /**
  * Classify the given text input, it returns the label of this input.
- * @param input the input text.
+ * @param _Input the input text.
  */
-export async function classify(input: string): Promise<string> {
+export async function classify(_Input: string): Promise<string> {
```
yorkie, line 8: Why rename this argument in this way?

packages/app/src/apis/vision.ts
```diff
@@ -3,9 +3,9 @@ import { dynamicModelExports } from './executable';
 
 /**
  * Classifies the image with given label, it returns the classification of this image.
- * @param img the input image type.
+ * @param _img the input image type.
  */
-export async function classify(img: Image): Promise<string> {
+export async function classify(_img: Image): Promise<string> {
```
yorkie, line 8: I think we can disable the unused check for this directory instead of renaming this argument.

packages/app/src/compiler/index.ts
```diff
@@ -7,7 +7,7 @@ export async function compile(pathname: string, tsconfig: string): Promise<Pipel
   const project = new Project({ tsConfigFilePath: tsconfig });
   const script = project.getSourceFileOrThrow(pathname);
   // TODO: needs to verify if this imported the learnable.
-  const app = script.getImportDeclarationOrThrow('@pipcook/app');
+  const _app = script.getImportDeclarationOrThrow('@pipcook/app');
```
yorkie, line 10: We can remove the `app` declaration.

packages/plugins/model-evaluate/fasttext-model-evaluate/src/index.ts
```diff
@@ -1,10 +1,9 @@
 import { ModelEvaluateType, UniModel, CsvDataset, ArgsType, EvaluateResult } from '@pipcook/pipcook-core';
-const boa = require('@pipcook/boa');
 
 const evaluate: ModelEvaluateType = async (dataset: CsvDataset, model: UniModel, args: ArgsType): Promise<EvaluateResult> => {
   const { expect = 0.5 } = args;
   if (dataset.testCsvPath) {
-    const [ samples, precision, recall ] = model.model.test(dataset.testCsvPath);
+    const [ _amples, precision, recall ] = model.model.test(dataset.testCsvPath);
```
yorkie, line 6:
```suggestion
    const [ _samples, precision, recall ] = model.model.test(dataset.testCsvPath);
```

packages/plugins/model-evaluate/image-generation-tensorflow-model-evaluate/src/index.ts
```diff
@@ -1,7 +1,7 @@
 import { ModelEvaluateType, UniModel, ImageDataset, EvaluateResult, ArgsType } from '@pipcook/pipcook-core';
 
 const cycleGanModelEvaluate: ModelEvaluateType
-  = async (data: ImageDataset, model: UniModel, args: ArgsType): Promise<EvaluateResult> => {
+  = async (data: ImageDataset, model: UniModel, _args: ArgsType): Promise<EvaluateResult> => {
```
yorkie, line 4: We should support the `ModelEvaluateType` could be compatible with `(data: ImageDataset, model: UniModel) => Promise<EvaluateResult>` instead of marking this to be ignored by lint tools.

packages/app/src/apis/nlp.ts
FeelyChau: Fixed to `_input`.

packages/app/src/apis/vision.ts
yorkie: This names would be in our API reference, and those are weird IMHO :(

packages/app/src/compiler/index.ts
FeelyChau: Should the todo comment be removed together?

FeelyChau:
> I'm -1 on this way to ignore this rule because this introduces an extra naming restriction, maybe we need to fix them correctly or add an eslint comment to disable some.

Prefixes are much simpler than eslint annotations, if they have to be defined.

yorkie:
> > I'm -1 on this way to ignore this rule because this introduces an extra naming restriction, maybe we need to fix them correctly or add an eslint comment to disable some.
> 
> Prefixes are much simpler than eslint annotations, if they have to be defined.

This added extra knowledge about writing pipcook.

yorkie:
And we need to find why do we use those unused variables, instead of allowing a hidden rule to ignore those.

packages/app/src/compiler/index.ts
yorkie: Just left the TODO

packages/app/src/compiler/index.ts
```diff
@@ -7,7 +7,6 @@ export async function compile(pathname: string, tsconfig: string): Promise<Pipel
   const project = new Project({ tsConfigFilePath: tsconfig });
   const script = project.getSourceFileOrThrow(pathname);
   // TODO: needs to verify if this imported the learnable.
-  const app = script.getImportDeclarationOrThrow('@pipcook/app');
```
yorkie, line 10: Sorry, what I did mean is to remove the constant `app`, and we need calling `getImportDeclarationOrThrow` to ensure the input source importing `@pipcook/app`.

packages/boa/.eslintrc.js
```diff
@@ -46,6 +46,7 @@ module.exports = {
     "capitalized-comments": "off",
     "class-methods-use-this": "error",
     "comma-dangle": "off",
+    '@typescript-eslint/no-unused-vars': 'off',
```
yorkie, line 49: I think we don't need this because the typescript rule won't have effect on JavaScript files.

FeelyChau: When I run `npm run lint` for boa project, the same error will be detected by both `@typescript-eslint/no-unused-vars` and `no-unused-vars`.

FeelyChau:
```shell
$ npm run lint
> @pipcook/boa@1.0.3 lint /Users/x/pipcook/packages/boa
> eslint . && ./clang-format.py

/Users/x/pipcook/packages/boa/tests/base/basic.js
  131:12  warning  '_' is assigned a value but never used  @typescript-eslint/no-unused-vars
```

yorkie: How about removing the `_` as the warning tip?

FeelyChau: I mean the plugin `@typescript-eslint/no-unused-vars` was applied to the js code.
And I think this test case can't be fixed if we need to check the array deconstruction of the builtin object. 
```js
t.throws(() => {
    // eslint-disable-next-line no-unused-vars
    const [_] = builtins;
    // Should throw the error
  }, TypeError);
```

yorkie: We can just print the destructed constants via:

```sh
console.log(_);
```

.eslintrc.js
```diff
@@ -45,6 +46,7 @@ module.exports = {
     'arrow-spacing': [ 'error', { before: true, after: true } ],
     'block-spacing': 'error',
     'brace-style': [ 'error', '1tbs', { allowSingleLine: true } ],
-    'comma-style': 'error'
+    'comma-style': 'error',
+    "no-unused-vars": [ "off" ], // Use @typescript-eslint/no-unused-vars instead, otherwise there will be checked twice
```
yorkie, line 50:
```suggestion
    'no-unused-vars': [ 'off' ], // Use @typescript-eslint/no-unused-vars instead, otherwise there will be checked twice
```

packages/plugins/model-evaluate/image-generation-tensorflow-model-evaluate/src/index.ts
```diff
@@ -1,7 +1,7 @@
-import { ModelEvaluateType, UniModel, ImageDataset, EvaluateResult, ArgsType } from '@pipcook/pipcook-core';
+import { ModelEvaluateType, UniModel, ImageDataset, EvaluateResult } from '@pipcook/pipcook-core';
 
 const cycleGanModelEvaluate: ModelEvaluateType
-  = async (data: ImageDataset, model: UniModel, args: ArgsType): Promise<EvaluateResult> => {
+  = async (data: ImageDataset, model: UniModel): Promise<EvaluateResult> => {
```
yorkie, line 4: This is not compatible with the type [`ModelEvaluateType`](https://github.com/alibaba/pipcook/blob/master/packages/core/src/types/plugins.ts#L176-L178) yet, we need some updates on the interface.

yorkie: @FeelyChau May I ask you to fix it in another pull request?

FeelyChau: OK, I'll change the `args` to optional.

FeelyChau: I checked the document of ts function here:
https://www.typescriptlang.org/docs/handbook/type-compatibility.html#comparing-two-functions
as the doc said, it's valid to pass `(data: ImageDataset, model: UniModel): Promise<EvaluateResult>` to `(data: ImageDataset, model: UniModel, args: ArgsType): Promise<EvaluateResult>`.

yorkie: Do we have a way to make this fail?

FeelyChau: This is valid in ts, and very useful, like the example in the document:
```js
let items = [1, 2, 3];

// Don't force these extra parameters
items.forEach((item, index, array) => console.log(item));

// Should be OK!
items.forEach(item => console.log(item));
```
If we make this fail, the case above will be failed too.

yorkie:
Note: please ignore the `gradientci` status!

rickycao-qy:
LGTM

Issue #362, opened 2020-07-23
no template for modelDefine plugin


WenheLI:
When running the following commands to generate a `modelDefine` plugin template:
```bash
pipcook plugin-dev --type modelDefine --name  tensorflow-rnn-model-define
```
the system will throw the following error:
```bash
Error: ENOENT: no such file or directory, copyfile '/usr/local/lib/node_modules/@pipcook/pipcook-cli/dist/assets/pluginPackage/src/modelDefine.ts'
```
suggesting the template for `modelDefine` does not exist.

yorkie:
Thanks for the catch, BTW we have another plan to use the `npm-init` project instead of self-maintained templates at https://github.com/alibaba/pipcook/issues/154.

WenheLI:
closed by #363

PR #363, opened 2020-07-23
cli: add modelDefine template

WenheLI:
This PR is to address #362 that adds a template for `modelDefine` plugin.

### DIFF
--- a/packages/cli/assets/pluginPackage/src/modelDefine.ts
+++ b/packages/cli/assets/pluginPackage/src/modelDefine.ts
@@ -1,23 +1,23 @@
 import {
-    ModelDefineType,
-    UniDataset,
-    ModelDefineArgsType,
-    UniModel,
-    Sample
+  ModelDefineType,
+  UniDataset,
+  ModelDefineArgsType,
+  UniModel,
+  Sample
 } from '@pipcook/pipcook-core';
 
 const templateModelDefine: ModelDefineType = async (data: UniDataset, args: ModelDefineArgsType): Promise <UniModel> => {
 
-    let model = {}; // Some model defination
+  let model = {}; // Some model defination
 
-    const retModelDefine: UniModel = {
-        model,
-        predict: async function (input: Sample) {
-            const pred = this.model.predict(input.data);
-            return pred.toString();
-        }
-    };
-    return retModelDefine;
+  const retModelDefine: UniModel = {
+    model,
+    predict: async function (input: Sample) {
+      const pred = this.model.predict(input.data);
+      return pred.toString();
+    }
+  };
+  return retModelDefine;
 };
 
 export default templateModelDefine;
--- a/packages/cli/assets/pluginPackage/src/modelDefine.ts
+++ b/packages/cli/assets/pluginPackage/src/modelDefine.ts
@@ -1,0 +1,23 @@
+import {
+    ModelDefineType,
+    UniDataset,
+    ModelDefineArgsType,
+    UniModel,
+    Sample
+} from "@pipcook/pipcook-core";
+
+const templateModelDefine: ModelDefineType = async (data: UniDataset, args: ModelDefineArgsType): Promise <UniModel> => {
+
+    let model = {}; // Some model defination
+
+    const retModelDefine: UniModel = {
+        model,
+        predict: async function (input: Sample) {
+            const pred = this.model.predict(input.data);
+            return pred.toString();
+        }
+    };
+    return retModelDefine;
+};
+
+export default templateModelDefine;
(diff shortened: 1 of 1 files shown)


packages/cli/assets/pluginPackage/src/modelDefine.ts
```diff
@@ -0,0 +1,23 @@
+import {
+    ModelDefineType,
+    UniDataset,
+    ModelDefineArgsType,
+    UniModel,
+    Sample
+} from "@pipcook/pipcook-core";
```
yorkie, line 7:
```suggestion
} from '@pipcook/pipcook-core';
```

yorkie:
The indention should be 2 spaces.

WenheLI:
> The indention should be 2 spaces.

Done

packages/cli/assets/pluginPackage/src/modelDefine.ts
```diff
@@ -0,0 +1,23 @@
...
+  ModelDefineType,
+  UniDataset,
+  ModelDefineArgsType,
+  UniModel,
+  Sample
+} from '@pipcook/pipcook-core';
+
+const templateModelDefine: ModelDefineType = async (data: UniDataset, args: ModelDefineArgsType): Promise <UniModel> => {
+
+  let model = {}; // Some model defination
```
FeelyChau, line 11:
```suggestion
  const model = {}; // Some model defination
```

packages/cli/assets/pluginPackage/src/modelDefine.ts
```diff
@@ -0,0 +1,23 @@
...
+
+  const retModelDefine: UniModel = {
+    model,
+    predict: async function (input: Sample) {
```
FeelyChau, line 15:
```suggestion
    predict: async (input: Sample) => {
```

FeelyChau: Should bind this.

yorkie: It's not necessary to use an arrow function here, the `this` is a reference to `retModelDefine`, which is correct.

WenheLI: +1; I think the two forms here are identical.

FeelyChau: Yeah, it' correct.

packages/cli/assets/pluginPackage/src/modelDefine.ts
```diff
@@ -0,0 +1,23 @@
...
+  const model = {}; // Some model defination
+
+  const retModelDefine: UniModel = {
+    model,
+    predict: async function (input: Sample) {
+      const pred = this.model.predict(input.data);
+      return pred.toString();
+    }
+  };
+  return retModelDefine;
```
FeelyChau, line 20: Not a Promise object.

packages/cli/assets/pluginPackage/src/modelDefine.ts
```diff
@@ -0,0 +1,23 @@
+import {
+  ModelDefineType,
+  UniDataset,
+  ModelDefineArgsType,
+  UniModel,
+  Sample
+} from '@pipcook/pipcook-core';
+
+const templateModelDefine: ModelDefineType = async (data: UniDataset, args: ModelDefineArgsType): Promise <UniModel> => {
```
FeelyChau, line 9: `async` is not needed, or use `await` in this function body.

packages/cli/assets/pluginPackage/src/modelDefine.ts
```diff
@@ -0,0 +1,23 @@
...
+
+  const model = {}; // Some model defination
+
+  const retModelDefine: UniModel = {
+    model,
+    predict: async function (input: Sample) {
```
FeelyChau, line 15: Ditto, remove `async`, or
```js
const pred = await this.model.predict(input.data);
```

packages/cli/assets/pluginPackage/src/modelDefine.ts
WenheLI: Because of the outer `async`(line 9), this object is indeed a Promise Object.

packages/cli/assets/pluginPackage/src/modelDefine.ts
WenheLI: Not sure if model.predict is an async function since this is binding from Python. Any references here?

FeelyChau: So remove `async` is more strict.

FeelyChau: So remove async is more strict.

FeelyChau: OK, I checked the `ModelDefineType`. You are correct.

PR #369, opened 2020-07-24
CI/CD: add nightly release

WenheLI:
This PR adds a new workflow that allows developers push a nightly build version to npm.

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -2,7 +2,7 @@ name: Nightly Build Packages
 on:
   push:
     tags:
-      - 'nightly'
+      - '**alpha**'
 
 jobs:
   release:
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -1,0 +1,34 @@
+name: Nightly Build Packages
+on:
+  push:
+    tags:
+      - 'nightly'
+
+jobs:
+  release:
+    name: Release
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Repo
+        uses: actions/checkout@v2
+
+      - name: Setup Node.js
+        uses: actions/setup-node@v1
+        with:
+          node-version: 12
+
+      - name: Install Dependencies
+        run: |
+          npm install
+          npm run build
+
+      - name: Publish npm nightly packages
+        run: |
+          echo "//registry.npmjs.org/:_authToken=${{ secrets.npm_token }}" > ~/.npmrc
+          git config --global user.name 'pipcook'
+          git config --global user.email 'queyue.crk@alibaba-inc.com'
+          npm run nightly-release
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          NODE_AUTH_TOKEN: ${{secrets.npm_token}}
+          NPM_AUTH_TOKEN: ${{ secrets.npm_token }}
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -1,8 +1,7 @@
 name: Nightly Build Packages
 on:
-  push:
-    branches:
-      - 'master'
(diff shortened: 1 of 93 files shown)


rickycao-qy:
for nightly-release script, since we use position parameter 'from-package', learn will compare version in package.json and npm registry and only publish those versions not existing in the registry yet. In my opinion, this command will not publish anything. Is it better to use something like 'lerna publish prerelease --force-publish --yes --no-verify-access --dist-tag nightly' ?

rickycao-qy:
And do we need to use a separate branch to do this?  Right now it seems to push all pre-releases version in package.json into master

WenheLI:
@rickycao-qy - Yep, I think we should maintain a separate branch `canary` or `nightly` to do the nightly release work. I will modify this script later.

.github/workflows/nightly.yml
```diff
@@ -0,0 +1,38 @@
+name: Nightly Build Packages
+on:
+  push:
+    tags:
+      - '**alpha**'
```
yorkie, line 5: It means that the nightly release needs someone to do publish tags manually.

yorkie:
@WenheLI @rickycao-qy And we need to document what's the Pipcook nightly build and how to use it at gitbook and [project guide](https://github.com/alibaba/pipcook/blob/master/docs/meta/PROJECT_GUIDE.md#release-lifecycle).

.github/workflows/nightly.yml
WenheLI: @yorkie - If we want to trigger this workflow every time master changed automatically, we need to delete & re-create the nightly branch to avoid conflict. You can refer to the new commits. Not sure if this is the proper way of implementing.

yorkie: This would add no necessary releases because we have many commits in 1 day, how about creating daily job via cron, and check if the `master` tree is updated in the workflow?

WenheLI:
Update: Use `cron` to trigger this workflow every day & check if the last commit is made within one day to decide whether we should publish a new nightly release version.

.github/workflows/nightly.yml
```diff
@@ -0,0 +1,66 @@
...
+        uses: actions/checkout@v2
+        with:
+          ref: nightly
+      - name:  Activity check
+        run: |
+            curl -sL https://api.github.com/repos/${{github.repository}}/commits | jq -r '[.[]' > $HOME/commit.json
+            date="$(jq -r '.commit.author.date' $HOME/commit.json)"
+            timestamp=$(date --utc -d "$date" +%s)
+
+            days=$(( ( $(date --utc +%s) - $timestamp ) / 86400 ))
```
yorkie, line 34: BTW, the GitHub provides a parameter `since` for filtering the commits by date, just curious if there is any other reason to write the time filter in bash :)

yorkie: > BTW, the GitHub provides a parameter `since` for filtering the commits by date, just curious if there is any other reason to write the time filter in bash :)

https://api.github.com/repos/alibaba/pipcook/commits?since=yesterday works! It looks easier to read :p

.github/workflows/nightly.yml
```diff
@@ -0,0 +1,66 @@
...
+        with:
+          branch: 'nightly'
+
+      - name: Checkout nightly
+        uses: actions/checkout@v2
+        with:
+          ref: nightly
+      - name:  Activity check
+        run: |
+            curl -sL https://api.github.com/repos/${{github.repository}}/commits | jq -r '[.[]' > $HOME/commit.json
```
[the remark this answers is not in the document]
WenheLI: Yep
From the official [doc](https://docs.github.com/en/rest/reference/repos#commits)
>  Default: the repository’s default branch (usually master).

WenheLI:
Too many conflicts to resolve while doing rebase, I will open another PR to fix it.

PR #373, opened 2020-07-27
daemon: throw error if the package is a local path and the pipeline config is from http/https url

FeelyChau:
Check the package of pipeline config which is from 'http/https' protocol.


rickycao-qy:
LGTM

PR #375, opened 2020-07-27
pipboard: fix the lint crash by eslint-plugin-react

yorkie:
The eslint-plugin-react@7.20.4 introduced a crash bug when linting a source with an anonymous function with the ES modules. This PR removes all related anonymous functions to hotfix the problem.

See https://github.com/yannickcr/eslint-plugin-react/issues/2728 for more information.


rickycao-qy:
LGTM

Issue #249, opened 2020-06-18
daemon: how do we handle with the database migration


yorkie:
Currently, we have no ways to help users to migrate their DB data when upgrading Pipcook, and we need to mark a pull request as `semver-major` when it tries to change the database schema.

FeelyChau:
Not applicable to Pipcook V2. This issue could be closed.

PR #381, opened 2020-07-28
plugin: add GlobalDataProcess Plugin

WenheLI:
This PR adds a new type of plugin that allows users to process the whole dataset.


packages/cli/assets/pluginPackage/src/dataProcess.ts
```diff
@@ -1,7 +1,6 @@
-import { DataProcessType, ArgsType, UniDataset } from '@pipcook/pipcook-core'
+import { DataProcessType, ArgsType, Sample } from '@pipcook/pipcook-core'
 
-const templateDataProcess: DataProcessType = async (data: UniDataset, args: ArgsType): Promise<UniDataset> => {
-  return {} as UniDataset;
+const templateDataProcess: DataProcessType = async (data: Sample, args: ArgsType): Promise<void> => {
```
rickycao-qy, line 3: Does the local data-process type still have metadata for the second parameter? Cuz it still has the responsibility to change metadata as discussed

yorkie:
The GlobalDataProcess is too long and implicit IMO, shall we have a better name for that?

WenheLI:
@yorkie - How about `DatasetProcess` plugin?

yorkie:
> @yorkie - How about `DatasetProcess` plugin?

I'm indefinitely +1 on this :p

packages/cli/assets/pluginPackage/src/globalDataProcess.ts
```diff
@@ -0,0 +1,6 @@
+import { DatasetProcessType, ArgsType, UniDataset } from '@pipcook/pipcook-core'
+
+const templateDatasetProcess: DatasetProcessType = async (data: UniDataset, args: ArgsType): Promise<void> => {
+}
```
yorkie, line 4:
```suggestion
};
```

packages/cli/assets/pluginPackage/src/globalDataProcess.ts
```diff
@@ -0,0 +1,6 @@
+import { DatasetProcessType, ArgsType, UniDataset } from '@pipcook/pipcook-core'
+
+const templateDatasetProcess: DatasetProcessType = async (data: UniDataset, args: ArgsType): Promise<void> => {
```
yorkie, line 3: Would you please add a comment line?

yorkie:
@WenheLI we upgraded the Pipeline Schema in this PR, does that mean it's incompatible with the current db storage?

WenheLI:
@yorkie - Nice catch! It seems `sequelize` does not provide auto-migrations. I think we need to implement [Migration](https://sequelize.org/master/manual/migrations.html) before merging this PR.

yorkie:
It's related with #249, and I have a question about where to store the migration scripts? @WenheLI

WenheLI:
@yorkie - Add a folder under `Daemon` called `scripts` or `migration` to store them might be appropriate?

yorkie:
> @yorkie - Add a folder under `Daemon` called `scripts` or `migration` to store them might be appropriate?

Ok, then they will be executed when someone upgrades a newer daemon?

WenheLI:
> > @yorkie - Add a folder under `Daemon` called `scripts` or `migration` to store them might be appropriate?
> 
> Ok, then they will be executed when someone upgrades a newer daemon?

Can we put a value like `DB_VERSION` in the database and `package.json`? If there is any updates regarding table structure, we manually update the version in `package.json`. And every time we start the daemon if the two versions are inconsistent, we run the script.

yorkie:
Totally agreed, and we can remove [`specVersion`](https://github.com/alibaba/pipcook/blob/master/packages/daemon/src/model/job.ts#L47-L51) after it gets landed :)

rickycao-qy:
LGTM once migration DB is ready

rickycao-qy:
And please resolve confclits :)

yorkie:
By the way, @WenheLI do we need to add a migration script for this PR?

WenheLI:
> By the way, @WenheLI do we need to add a migration script for this PR?

The migration script is added by another PR(#397 )

PR #383, opened 2020-07-29
boa: use the specific version to fix build failure

yorkie:
The latest miniconda has switched to Python 3.8 :)


yorkie:
The failures of cli are caused by it used to the NPM registry to install which is crashed, either. So that means we need a hotfix release ASAP.

Issue #314, opened 2020-07-07
cli: failed to run `pipcook daemon start` on my macOS


ivc369:
### Operating system and NodeJS version(s):
* MacOS Catalina `10.15.5`
* NodeJS `v12.18.2`

### Pipcook version(s):
`v1.0.1`

I ran `pipcook daemon start` on MacOS and encountered the following error: 

![image](https://user-images.githubusercontent.com/16490377/86733300-aaea5280-c063-11ea-8a7c-9b13fb104c04.png)

yorkie:
@ivc369 Thanks for your feedback, I have investigated this problem a bit, it might be caused by a failed `pipcook init`, then the daemon directory doesn't get created. However the Node.js child process's `fork()` will throw the ENOENT by calling the [`chdir(2)`](https://github.com/libuv/libuv/blob/v1.x/src/unix/process.c#L341-L344) with non-existed `cwd` path.

I think we could check the daemon related paths are set up correctly and give a better tip before executing the `pipcook-daemon(1)` commands.

/cc @FeelyChau

FeelyChau:
@ivc369 it looks like the daemon is not installed successfully.
Could you re-run `pipcook init`, then try again?

ivc369:
@FeelyChau I tried to run `sudo pipcook init` again, and then an exception occurred:

```bash
exec ""npm" install @pipcook/daemon -E --production"
npm WARN deprecated request@2.88.2: request has been deprecated, see https://github.com/request/request/issues/3142

> @pipcook/boa@1.0.1 preinstall /Users/xxx/.pipcook/server/node_modules/@pipcook/boa
> node tools/check-dependence.js && make -C ./pybind11/ && node tools/install-python.js && node tools/install-requirements.js

/bin/rm
/usr/bin/make
/usr/bin/tar
/usr/bin/curl
make: Nothing to be done for `all'.
internal/fs/utils.js:230
    throw err;
    ^

Error: EACCES: permission denied, open '/Users/xxx/.pipcook/server/node_modules/@pipcook/boa/.CONDA_INSTALL_DIR'
    at Object.openSync (fs.js:458:3)
    at Object.writeFileSync (fs.js:1355:35)
    at exports.initAndGetCondaPath (/Users/xxx/.pipcook/server/node_modules/@pipcook/boa/tools/utils.js:31:6)
[...]

PR #384, opened 2020-07-29
notebook: remove `sudo` in the code cell

ivc369:
When running `!sudo pipcook init` and `!sudo pipcook daemon start` in my Notebook, this error ocurred:

```bash
/bin/sh: 1: sudo: not found
/bin/sh: 1: sudo: not found
```

And I was previously told to run `pipcook init` without `sudo`. [https://github.com/alibaba/pipcook/issues/314](https://github.com/alibaba/pipcook/issues/314)

### DIFF
--- a/notebooks/pipcook_image_classification.ipynb
+++ b/notebooks/pipcook_image_classification.ipynb
@@ -97,8 +97,8 @@
         "colab": {}
       },
       "source": [
-        "!sudo pipcook init\n",
-        "!sudo pipcook daemon start"
+        "!pipcook init\n",
+        "!pipcook daemon start"
       ],
       "execution_count": null,
       "outputs": []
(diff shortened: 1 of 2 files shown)


yorkie:
@ivc369 Are you running at Google Colab?

ivc369:
> @ivc369 Are you running at Google Colab?

No I use self-built JupyterLab..

yorkie:
@rickycao-qy Do you know if we can remove the `sudo` at Google Colab, too?

rickycao-qy:
@ivc369 the notebooks in notebook folder are prepared just for Colab. For a normal environment, we are suggesting to use normal process as instructed in Readme. @yorkie Or we can rename the folder to colab instead of notebook

yorkie:
> @ivc369 the notebooks in notebook folder are prepared just for Colab. For a normal environment, we are suggesting to use normal process as instructed in Readme. @yorkie Or we can rename the folder to colab instead of notebook

I think a note in README is better, which describes what jupyter notebooks are supported and why do we choose them :)

ivc369:
I got it, thanks :)

PR #388, opened 2020-07-30
costa: add python runnable for machine-learning developers

yorkie:
In order to better attract machine learning developers to contribute models to Pipcook, this PR provides developers with a native Python development environment. For users who do not use Node.js and JavaScript, they can use pure Python to write plugins.

### DIFF
--- a/docs/contributing/contribute-a-plugin.md
+++ b/docs/contributing/contribute-a-plugin.md
@@ -25,27 +25,7 @@ const collectTextline: DataCollectType = async (args: ArgsType): Promise<void> =
 For other plugin interfaces, see [this list of plugin categories](../spec/plugin.md#plugin-category).
 
 
-## Test your plugin
-
-You can prepare the corresponding mocking data based on the interface. For example, if you developed a `DataProcessType` plugin, create some mock firstly:
-
-```js
-const data = {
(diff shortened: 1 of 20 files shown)


yorkie:
Tests and documentation are still working in progress :)

rickycao-qy:
To get this working, Boa needs to support pass Javascript (class, function, etc.) to Python and executed in Python. Otherwise, this will not work.

yorkie:
@rickycao-qy Which case do we need to pass a function/class to Python?

yorkie:
@rickycao-qy I have updated the tests for using JavaScript functions in Python env at packages/costa/test/plugins/python-simple/__init__.py, the only problem is that the Python objects from JavaScript are a dictionary so that:

```py
def main(input):
  input.get('fn1')() # working
  input.fn1() # not working
```

I will open another PR to change to case 2.

yorkie:
@rickycao-qy I have supported access to the object attributes in dot notation(#389), I will switch to dot when it gets merged like:

```py
def main(input):
  input.fn1()
```

yorkie:
@rickycao-qy @FeelyChau I have rebased this pull request :)

rickycao-qy:
LGTM

PR #389, opened 2020-07-31
boa: generate object instead of `dict` to Python function

yorkie:
It creates a new class and instantiates it with given props when passing to Python. For example:

```js
pythonfunc({ foobar: 10 });
```

At Python, we have to write in this way:

```py
def func(opt):
  opt.get('foobar') # 10, because opt is a dict
```

Within this Pull Request, the developer is able:

```py
def func(opt):
  opt.foobar
```

Otherwise, I also removed an incorrect case:

```js
const o = { foobar: 10 };
len(o); // not working
len(dict(boa.kwargs(o))); // working
```

Even though it's more complexed than before, but it's more correct way to be used.

### DIFF
--- a/packages/boa/lib/index.js
+++ b/packages/boa/lib/index.js
@@ -196,7 +196,9 @@ function _internalWrap(T, src={}) {
     invoke: {
       enumerable: false,
       writable: false,
-      value: args => T.invoke.apply(T, args),
+      value: args => {
+        return T.invoke.apply(T, args);
+      },
     },
     /**
      * @method toString
--- a/packages/boa/src/core/object.cc
+++ b/packages/boa/src/core/object.cc
@@ -422,7 +422,6 @@ PyObject *PythonObject::Cast(Napi::Env env, Object value,
                              bool finalizeFuncType) {
   auto names = value.GetPropertyNames();
   auto dict = PyDict_New();
-  auto isKwargs = IsKwargs(value);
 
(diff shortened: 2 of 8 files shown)


packages/boa/lib/index.js
```diff
@@ -196,7 +196,9 @@ function _internalWrap(T, src={}) {
     invoke: {
       enumerable: false,
       writable: false,
-      value: args => T.invoke.apply(T, args),
+      value: args => {
+        return T.invoke.apply(T, args);
```
rickycao-qy, line 200: What's the difference

yorkie:
Could you please describe what the problem is? And it is better to provide a test case to reproduce.

packages/boa/lib/index.js
yorkie: No any differences, just add brackets to wrap the function.

FeelyChau:
> Could you please describe what the problem is? And it is better to provide a test case to reproduce.

In this case, the class `CycleGan` constructor expects `dict`, it will throw error if we merge this PR, so we need to refactor them together.

yorkie:
> > Could you please describe what the problem is? And it is better to provide a test case to reproduce.
> 
> In this case, the class `CycleGan` constructor expects `dict`, it will throw error if we merge this PR, so we need to refactor them together.

Ok, I will wrap them with a `dict()`.

rickycao-qy:
LGTM

PR #392, opened 2020-07-31
build: tweak the versions to decrease root packages' size

### LINKED ISSUE #344: build: update versions to decrease the size of root node_module
I investigated some of our dependencies trees and found the following optimizable actions to decrease **~700MB** disk space of our node_modules:

### `@tensorflow` packages (~100MB)

Inconsistent `@tensorflow/tfjs*`, which saves many copies of these packages, and we are able to decrease by 100MB by using the same version of these packages:

```
53848	./node_modules//@tensorflow/tfjs-node-gpu/node_modules/@tensorflow/tfjs-core
34808	./node_modules//@tensorflow/tfjs-node-gpu/node_modules/@tensorflow/tfjs
22024	./node_modules//@tensorflow/tfjs-node-gpu/node_modules/@tensorflow/tfjs-layers
8176	./node_modules//@tensorflow/tfjs-node-gpu/node_modules/@tensorflow/tfjs-converter
4744	./node_modules//@tensorflow/tfjs-node-gpu/node_modules/@tensorflow/tfjs-data
```

### `@jimp` and `core-js` (~400MB)

`@jimp` packages have a common dependency `core-js@3.6.x` which takes 15MB, unfortunately, there are many `@jimp` packages also install the same package separately, and that takes up to 400MB in our node_modules. The reason that the lerna repeats to install the same `core-js` package is that the babel-runtime uses a downgraded version `core-js@2.4.x` which is incompatible with the former, so the lerna will install all `core-js` packages independently.

To resolve the problem, we have 2 alternative solutions:

1. downgrade the `jimp` packages to compatible with `babel-runtime`.
2. upgrade the `babel-runtime`, but we still need to find which are the dependents of this package.

### `typescript` (~200MB)

We have the same issue on `typescript` dependents:

- `@ts-morph/common` uses the latest `typescript@3.9.x`.
- `typedoc` uses the `typescript@3.7.x`.
- root project uses the `typescript@3.7.x`, too.

So every `typescript` would be installed independently, and every installation takes ~100MB.

yorkie:
This fixes #344, and the `node_modules` result is:

| before | after |
|--------|------|
| 1.3G | 826M |

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -19,6 +19,12 @@
     },
     "bootstrap": {
       "nohoist": [
+        "@alifd/next",
+        "@icedesign/container",
+        "@icedesign/img",
+        "@icedesign/layout",
+        "@icedesign/notification",
+        "@icedesign/theme",
         "node-addon-api",
         "midway",
         "egg-ci",
--- a/package-lock.json
+++ b/package-lock.json
@@ -1641,7 +1641,8 @@
     "@types/jasmine": {
       "version": "3.5.10",
       "resolved": "https://registry.npmjs.org/@types/jasmine/-/jasmine-3.5.10.tgz",
-      "integrity": "sha512-3F8qpwBAiVc5+HPJeXJpbrl+XjawGmciN5LgiO7Gv1pl1RHtjoMNqZpqEksaPJW05ViKe8snYInRs6xB25Xdew=="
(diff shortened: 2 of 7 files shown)


rickycao-qy:
LGTM

FeelyChau:
There are some conflicts, could you resolve  them and commit again?

yorkie:
> There are some conflicts, could you resolve them and commit again?

Fixed, and let's wait for CI builds.

Issue #240, opened 2020-06-16
daemon: use randomBytes() instead of uuid


yorkie:
The current UUID is too long, and the UUID is not that necessary in our local environment.

FeelyChau:
RandomBytes is not strong enough, the collision probability increases with the increase of data.
There are some id-generators we can look at:
1. ~~git commit id: commit id is calculated based on the SHA-1 of the file content, but we don't have a file.~~
2. ( 8 bytes )[snowflake by twitter](https://github.com/twitter-archive/snowflake/releases/tag/snowflake-2010): It consists of timestamp (41bit), machine id (10bit), and serial number(12bit), and left 1 bit unused, but in our scenario, the machine id is determined.
3. ( 12 bytes )[mongodb object id](https://docs.mongodb.com/manual/reference/method/ObjectId/): It consists of timestamp (32bit), random value (40bit), and serial number(24bit), it looks nice to us but maybe too long.

so, I think we could implement a simplified generator like mongdb object id, but shortening data length.

yorkie:
> git commit id: commit id is calculated based on the SHA-1 of the file content, but we don't have a file.

Do we have the pipeline config file? how does the git compute the commit id by the file?

> mongodb object id: It consists of timestamp (32bit), random value (40bit), and serial number(24bit), it looks nice to us but maybe too long.

Yea, it's too long. How is docker generates an ID for its container and image?

FeelyChau:
> > git commit id: commit id is calculated based on the SHA-1 of the file content, but we don't have a file.
> 
> Do we have the pipeline config file? how does the git compute the commit id by the file?
> 
> > mongodb object id: It consists of timestamp (32bit), random value (40bit), and serial number(24bit), it looks nice to us but maybe too long.
> 
> Yea, it's too long. How is docker generates an ID for its container and image?

1. It's not a generic id-generator
2. [here is the code](https://github.com/docker/docker-ce/blob/6c81bbd59d78f35499d47a5832656b727e9925a0/components/engine/pkg/stringid/stringid.go#L40), it's too long for us.

FeelyChau:
Maybe we have another choice: use all the letters and numbers to compress the id, just like [this](https://www.npmjs.com/package/shortid)

yorkie:
> Maybe we have another choice: use all the letters and numbers to compress the id, just like [this](https://www.npmjs.com/package/shortid)

[...]

PR #394, opened 2020-08-02
core: replace shortid with nanoid

FeelyChau:
As marked in #240 , the package `shortid` is dead, we need to replace with [nanoid](https://github.com/ai/nanoid).


FeelyChau:
Because of the network, I need to push the drafts and run the CI. If it's ready, I'll remove the WIP mark.

FeelyChau:
Ready.

PR #399, opened 2020-08-03
build: re-installation takes too long

WenheLI:
This PR adds `clean` ahead of install process to avoid extremely long `prune hoisting ` time.
For more detail, please refer to this issue https://github.com/lerna/lerna/issues/602

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,6 +21,7 @@
     "typescript": "^3.9.7"
   },
   "scripts": {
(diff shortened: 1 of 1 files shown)


yorkie:
LGTM and that's great if we can provide some benchmarks, and I removed the `bug` label because we usually use that to file an issue is a bug that is to be addressed later.

rickycao-qy:
LGTM

PR #400, opened 2020-08-03
build, doc: remove deprecated typedoc-plugin-lerna

yorkie:
And we will generate our API documentation separately:

1. Pipcook SDK API
2. Pipcook Core API


rickycao-qy:
LGTM

PR #401, opened 2020-08-03
fix: reimplement cli whth sdk

FeelyChau:
These commands are ready:
* pipcook run
* pipcook start
* pipcook plugin
* pipcook pipeline
* pipcook job


FeelyChau:
If this PR is OK, we need to resolve conflicts with branch master and review all the codes again.

packages/cli/src/bin/pipcook-job.ts
```diff
@@ -1,40 +1,146 @@
...
+import { join, isAbsolute } from 'path';
+import * as fs from 'fs-extra';
+import tar from 'tar-stream';
+import { createGunzip } from 'zlib';
+import { PipelineStatus } from '@pipcook/pipcook-core';
+import { JobResp } from '@pipcook/sdk';
+import { tunaMirrorURI } from '../config';
+import { logger, initClient } from '../utils';
+
+const PipelineStatusStr = [ 'creating', 'running', 'success', 'fail' ];
```
rickycao-qy, line 13: Does this status constant also exist in daemon or core. Is it possible to just maintain one copy of this in core

packages/cli/src/bin/pipcook-job.ts
```diff
@@ -1,40 +1,146 @@
...
+
+async function list(opts: any): Promise<void> {
+  const client = initClient(opts.host, opts.port);
+  const jobs = client.job.list();
```
rickycao-qy, line 17: await client.job.list()?

packages/cli/src/bin/pipcook-job.ts
```diff
@@ -1,40 +1,146 @@
...
+async function runAndDownload (filename: string, opts: any) {
+  const job = await run(filename, opts);
+  const client = initClient(opts.host, opts.port);
+  const outputRootPath = join(process.cwd(), opts.output || 'output');
+  logger.info(`start to download output to ${outputRootPath}`);
+  // remove the output dir
+  await fs.remove(outputRootPath);
+  try {
+    // generate output
+    const fileDownloadResp = await client.job.downloadOutput(job.id);
```
rickycao-qy, line 119: Is it possbile to encapsulate logics of compress/uncompress into some public utils

packages/cli/src/bin/pipcook-job.ts
FeelyChau: For now, it's used in `cli` only, but I think moved into `core` is better, too.

FeelyChau: I checked the code again, it has beed defined in `sdk`. If it is used by client only, I think we can export it from `sdk`.

rickycao-qy: sounds good

packages/cli/src/bin/pipcook-job.ts
FeelyChau: Done.

packages/cli/src/bin/pipcook-job.ts
FeelyChau: Moved to the utils.ts, do you mean we should implement this in `core`?

rickycao-qy:
LGTM

PR #402, opened 2020-08-04
build: add beta preid to version

WenheLI:
This PR fixes the mismatch version issue

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,6 +31,6 @@
     "typedoc": "typedoc",
     "clean": "lerna run clean",
     "release": "lerna publish from-package --yes --no-verify-access",
-    "beta-release": "lerna version prerelease --force-publish=* --yes && lerna publish from-package  --preid beta --yes --no-verify-access --dist-tag beta"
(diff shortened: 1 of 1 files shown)


FeelyChau:
Because of the latest version is `1.0.6-alpha.0`, the unit test case is failed. 
https://github.com/alibaba/pipcook/blob/master/packages/costa/src/runtime_test.ts#L67
we need this PR to fix the latest version string on npm.

package.json
```diff
@@ -31,6 +31,6 @@
     "typedoc": "typedoc",
     "clean": "lerna run clean",
     "release": "lerna publish from-package --yes --no-verify-access",
-    "beta-release": "lerna version prerelease --force-publish=* --yes && lerna publish from-package  --preid beta --yes --no-verify-access --dist-tag beta"
+    "beta-release": "lerna version prerelease --preid beta --force-publish=* --yes && lerna publish from-package --preid beta --yes --no-verify-access --dist-tag beta"
```
rickycao-qy, line 34: --preid in lerna publish makes no effect

rickycao-qy:
LGTM

PR #405, opened 2020-08-04
Api refactor

FeelyChau:
Refactored:
* cli:
  * plugin: list, install, uninstall
  * pipeline: info, list, create, update, remove, install
  * job: list, run, remove, log, cancel(stop)
  * common: pipcook run
* daemon: all plugin, pipeline, job apis
* sdk: all plugin, pipeline, job apis
* test: test cases updated


packages/cli/src/bin/pipcook-job.ts
```diff
@@ -1,63 +1,153 @@
...
 import program from 'commander';
-import { get } from '../request';
-import { run } from '../pipeline';
-import { route } from '../router';
-import { logger } from '../utils';
+import { join, isAbsolute } from 'path';
+import * as fs from 'fs-extra';
+import { PipelineStatus } from '@pipcook/pipcook-core';
+import { JobStatus } from '@pipcook/sdk';
+import { JobResp } from '@pipcook/sdk';
```
yorkie, line 8: Duplicated imports for sdk.

packages/cli/src/bin/pipcook-job.ts
```diff
@@ -1,63 +1,153 @@
...
+    await extractToPath(fileDownloadResp.stream, outputRootPath);
+    logger.success('download finished');
+  } catch (err) {
+    logger.fail(err.message);
+  }
 }
 
 program
   .command('list')
+  .option('-h|--host <host>', 'the host of daemon')
```
yorkie, line 119: The host contains the port.

yorkie: And I think they would be in global config.

packages/costa/src/runnable.ts
```diff
@@ -59,15 +61,25 @@ export class PluginRunnable {
    */
   public state: 'init' | 'idle' | 'busy';
 
+  /**
+   * The flag somebody stop running
+   */
+  public cancelByUser: boolean;
```
yorkie, line 67:
```suggestion
  public canceled: boolean;
```

packages/costa/src/runnable.ts
```diff
@@ -1,8 +1,10 @@
 import path from 'path';
-import { ensureDir, ensureSymlink, ensureFile, open, close } from 'fs-extra';
+import { Writable } from 'stream';
```
yorkie, line 2: `NodeJS.WritableStream`.

packages/daemon/src/app/controller/base.ts
```diff
@@ -0,0 +1,68 @@
...
+   */
+  success(data?: any, status?: number): void {
+    this.ctx.body = data;
+    if (typeof status !== 'undefined') {
+      this.ctx.status = status;
+    } else {
+      this.ctx.status = typeof data === 'undefined' ? HttpStatus.NO_CONTENT : HttpStatus.OK;
+    }
+  }
+}
```
liangzr, line 35: How about extending a `success` method in `this.ctx`(inside `daemon/app/extend/context.ts`)? Then the usage of `this.ctx.success` and `this.ctx.throw` are more consistent.

Type definition could be extending globally, in `/path/to/type/midway.d.ts`:
```ts
declare module 'egg' {
  interface Context {
    success(): void
  }
}
```

FeelyChau: Nice! I'll update the BaseController.

packages/cli/src/bin/pipcook-job.ts
FeelyChau: I think the better way is: the values in global config is default, users  also can use `-h` `-p` to custom the values for current operation.

packages/daemon/src/app/controller/base.ts
FeelyChau: I checked the issues for how to extend context, found this:
https://github.com/koajs/koa/issues/473
For now, I defined the `success` function in `bootstrap.js`. If there is any way  better, I'll work for it.

FeelyChau: I think I know what's your meaning.
https://eggjs.org/en/tutorials/progressive.html#prototype-of-plugin

tools/config.json
```diff
@@ -0,0 +1,14 @@
+[
```
yorkie, line 1: The filename is too general, we need a more specific one.

tools/init_dev.js
```diff
@@ -0,0 +1,74 @@
+const spawnSync = require('child_process').spawnSync;
```
yorkie, line 1: 'use strict' is missing

packages/cli/src/service/pipeline.ts
```diff
@@ -0,0 +1,140 @@
+
```
Txiaozhe, line 1: Invalid blank line:(

Txiaozhe:
LGTM but some nits and too many 'any' types:(

FeelyChau:
> LGTM but some nits and too many 'any' types:(

The 'any' in cli is because the opts object is formatted by `Commander`, so it's not so necessary.
But yes, we need to complete it.

packages/daemon/src/app/controller/base.ts
liangzr: Seems that I forgot to refer a link about extend context last time😂, sorry for that. You can refer to the following document supported by eggjs
- [Extend EGG](https://eggjs.org/en/basics/extend.html#access-method-2)

daemon/app/extend/context.ts:
```ts
import { Context } from 'midway'

export function success(this: Context, data?: any, status?: number) {
  // do something...
}
```

liangzr: @FeelyChau I noticed that you have extended it in context, but TypeScript can not understand framework conventions, we could extend the type definition globally:

```ts
declare module 'egg' {
  interface Context {
    success(data?: any, status?: number): void
  }
}
```

It looks a bit fragmented, but I think it’s a worthwhile improvement.

packages/daemon/src/service/pipeline.ts
```diff
@@ -273,25 +273,45 @@ export class PipelineService {
...
+        runnable.destroy();
+        delete runnableMap[id];
+      } else {
+        console.error(`no runnable found: ${id}`);
+      }
+      job.status = PipelineStatus.CANCELED;
+      await job.save();
     } else {
-      return false;
+      this.ctx.throw(HttpStatus.BAD_REQUEST, 'job is not running');
```
liangzr, line 314: Service shouldn't perceive the controller stuff, just throw an error to the enclosing controller

packages/costa/src/runnable.ts
```diff
@@ -63,15 +64,25 @@ export class PluginRunnable {
...
 
+  /**
+   * The flag somebody stop running
+   */
+  public canceled: boolean;
+
+  /**
+   * logger
+   */
+  private logger: Record<'stdout' | 'stderr', NodeJS.WritableStream>;
```
liangzr, line 75: Duplicated with existing interface `LogStdio`.

packages/costa/src/runtime.ts
```diff
@@ -394,13 +382,27 @@ export class CostaRuntime {
...
+    };
+    if (Array.isArray(name)) {
+      let success = false;
+      // any one uninstalls successfully, return true
+      await name.forEach(async (singleName) => {
+        if (await removePkg(singleName)) {
+          success = true;
+        }
+      });
+      return success;
```
liangzr, line 402: `await name.forEach` can not wait as expected and return false directly, try the following case:

```js
async function foo() {
    await [1].forEach(async () => {
        await new Promise(rsv => setTimeout(rsv, 1000))
        console.log('yes')
    })
    console.log('ok')
}
```

packages/daemon/src/service/pipeline.ts
FeelyChau: But how could throw a bad request error?

packages/costa/src/runnable.ts
FeelyChau: Done.

packages/daemon/src/service/pipeline.ts
liangzr: All right, we can't distinguish it with internal errors now without custom error, but it's not a good start since you can't find a typical response clearly in the controller strategy. Maybe this problem can be put together with the improvement of the global error handle at another PR.

FeelyChau: Maybe we can throw a [HttpError](https://www.npmjs.com/package/http-errors) which is the same as `ctx.throw` throw.

packages/daemon/src/app/controller/base.ts
FeelyChau: Done.

packages/daemon/src/typing/midway.d.ts
```diff
@@ -0,0 +1,7 @@
+
+declare module 'egg' {
+  interface Context {
+    success(data?: any, status?: number): void
+  }
+}
+
```
liangzr, line 7: - reuse the original comment at the type definition
- remove `space` char after the last newline
```suggestion

declare module 'egg' {
  interface Context {
    /**
     * send successful response, the status will be set to 204 or 200,
     * depend on if data is undefined
     * 
     * @param data data returned
     * @param status success code, should be [200, 299], default is 200
     */
    success(data?: any, status?: number): void
  }
}

```

Issue #364, opened 2020-07-23
test: support unit test for plugins


WenheLI:
We should support/add unit test for plugins to ensure the contributed plugins work well and leverage plugin developers' workload.
- For `data process`, `data collect`, and `data access`, we can just ensure with given input we can get wanted output.
- For the model related plugins, since their output is not definite and unpredictable, it is not easy/appropriate to operate the normal unite test. Any ideas on this part?

PR #406, opened 2020-08-04
test: add plugin level unit test

WenheLI:
To implement #364 , this adds two demo unit test for `csv` component.


yorkie:
Shall we move this to separated plugin repository?

FeelyChau:
Because we have move the plugins into [Imgcook](https://github.com/imgcook), so this PR should commit to the plugins repo.

PR #408, opened 2020-08-04
cli: fix ci error

FeelyChau:


.github/workflows/cli.yml
```diff
@@ -2,10 +2,10 @@ name: cli
 on:
   push:
     branches:
-    - master
+    - api-refactor
```
rickycao-qy, line 5: What's this branch used for

packages/cli/src/service/job.ts
```diff
@@ -0,0 +1,135 @@
...
+  }
+}
+
+export async function log(id: string, opts: any): Promise<void> {
+  const client = initClient(opts.host, opts.port);
+  try {
+    const log = await client.job.log(id);
+    console.log(log);
+  } catch (err) {
+    console.error(err.message);
```
rickycao-qy, line 53: suggest using logger object for all console-related outputs

.github/workflows/cli.yml
FeelyChau: Just for CI, need to roll back before merge.

packages/cli/src/actions/job.ts
```diff
@@ -0,0 +1,135 @@
...
+
+export async function list(opts: any): Promise<void> {
+  const client = initClient(opts.host, opts.port);
+  const jobs = await client.job.list();
+  if (jobs.length > 0) {
+    console.table(jobs.map((job) => {
+      return { ...job, status: JobStatusValue[job.status] };
+    }), [ 'id', 'status', 'evaluatePass', 'createdAt' ]);
+  } else {
+    console.info('no job is created.');
```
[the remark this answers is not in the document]
FeelyChau: done.

packages/cli/src/service/job.ts
FeelyChau: Done.

.github/workflows/cli.yml
yorkie: We can roll back because the CI passes.

tools/config.json
```diff
@@ -0,0 +1,5 @@
+{
+    "plugins": [ "boa", "core" ],
+    "server": [ "daemon" ],
+    "pipboard": [ "pipboard" ]
+}
```
yorkie, line 5: The `server` and `pipboard` are not that necessary to be configurable, and it adds extra learn-cost, I would prefer this config is only for plugins directory just like:

```json
[ "packages/boa", "packages/core" ]
```

And rename it to a more specific one just like "dependencies.json".

FeelyChau: If `daemon` and `pipboard` are not inited, the logic for test `isLocal` will be left.
I think we shouldn't put test codes in to the user-run codes.

FeelyChau: I modified the format to:
```json
[
  {
    "destDir": "plugins",
    "dependencies": [ "boa", "core" ]
  },
  {
    "destDir": "server",
    "dependencies": [ "daemon" ]
  },
  {
    "destDir": "pipboard",
    "dependencies": [ "pipboard" ]
  }
]

```
This should make it more readable.

FeelyChau: @rickycao-qy @WenheLI May I ask you to review this design?

rickycao-qy:
LGTM

tools/config.json
rickycao-qy: isLocal should be removed in the future

rickycao-qy: As i understand, this config is the configuration for something that will be pre-installed when init project. it's ok to leave anything included in this project

packages/daemon/src/migrations/20200803063219-addDatasetProcess.ts
```diff
@@ -0,0 +1,24 @@
+'use strict';
```
rickycao-qy, line 1: How does the daemon decide the execution order of this kind of migration script

tools/config.json
yorkie: What I meant is moving pipboard/daemon to init script instead of configuration.

packages/daemon/src/migrations/20200803063219-addDatasetProcess.ts
FeelyChau: Look at:
https://github.com/papb/tory/blob/ed1eb517d403390ba19ba7233bb881f6663cda84/source/helpers/sort-lexicographically.ts#L3

The order is lexicographically.
The require-link is [sequelize/cli](https://github.com/sequelize/cli/blob/master/src/core/migrator.js#L41)->[umzug](https://github.com/sequelize/umzug/blob/master/src/umzug.ts#L333)->[papb/tory](https://github.com/papb/tory/blob/ed1eb517d403390ba19ba7233bb881f6663cda84/source/helpers/sort-lexicographically.ts#L3).

PR #417, opened 2020-08-07
daemon: check table before migration

FeelyChau:
We should  check if the table exists before migration.

### DIFF
--- a/packages/daemon/src/migrations/20200803063219-addDatasetProcess.ts
+++ b/packages/daemon/src/migrations/20200803063219-addDatasetProcess.ts
@@ -20,7 +20,7 @@ export default {
 
   down: (queryInterface: QueryInterface) => {
     return queryInterface.sequelize.transaction(async t => {
-      const tbNames = await queryInterface.showAllTables();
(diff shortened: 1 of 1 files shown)


yorkie:
Shall we have tests for these migration scripts?

FeelyChau:
> Shall we have tests for these migration scripts?

I think it's necessary.

packages/daemon/src/migrations/20200803063219-addDatasetProcess.ts
```diff
@@ -3,22 +3,28 @@ import { QueryInterface, DataTypes } from 'sequelize';
...
+            type: DataTypes.STRING
+          }, { transaction: t })
+        ]);
+      }
     });
   },
 
   down: (queryInterface: QueryInterface) => {
     return queryInterface.sequelize.transaction(async t => {
+      const tbNames = await queryInterface.showAllTables();
```
yorkie, line 23:
```suggestion
      const tableNames = await queryInterface.showAllTables();
```

PR #418, opened 2020-08-07
deamon: run migration script when booting

WenheLI:


packages/daemon/bootstrap.js
```diff
@@ -5,6 +5,7 @@ const os = require('os');
 const { pathExists } = require('fs-extra');
 const fs = require('fs');
 const { start } = require('egg');
+const { exec } = require('child_process');
```
yorkie, line 8: Should we use execSync?

packages/daemon/bootstrap.js
```diff
@@ -41,6 +42,9 @@ function createPidfileSync(pathname) {
   // create pidfile firstly
   createPidfileSync(DAEMON_PIDFILE);
 
+  // run migration in sub-process
+  exec("npm run migration");
```
yorkie, line 46:
```suggestion
  exec('npm run migration');
```

FeelyChau, line 46: I think the down script should be executed if any error happens when migration.

WenheLI: If any error occurs during migration, no changes will be committed to the database. Therefore, in this case, I do not think we need to run the `down script`.

packages/daemon/bootstrap.js
FeelyChau: IMO, `sync` with `await` is better.

yorkie: What you meant is `async+await`? Actually we can't use this way in this use case because the top-level await has not supported.

FeelyChau: The `exec` is called in an async function now.

yorkie: This is why I added this comment, it might run the migration when a user is requesting something.

PR #421, opened 2020-08-09
test(boa): add dynamic import test on esm

rickyes:
Refs: https://nodejs.org/api/esm.html#esm_import_expressions

/cc @yorkie @FeelyChau


yorkie:
Shall we add a test for `await import()`?

rickyes:
> Shall we add a test for `await import()`?

I don't think it's necessary, it's just a difference in `syntactic sugar`, which is the same logic for `Pipcook`

PR #426, opened 2020-08-11
daemon: add fix for db migration

WenheLI:
This PR is to address migration failure.


packages/daemon/bootstrap.js
```diff
@@ -45,7 +45,9 @@ function createPidfileSync(pathname) {
 
   if (fs.existsSync(PIPCOOK_DB)) {
     // run migration in sub-process
-    execSync('npm run migration');
+    const res = execSync('npm run migration', {
+      "cwd": __dirname
```
yorkie, line 49:
```suggestion
      'cwd': __dirname
```

packages/daemon/src/migrations/20200803063219-addDatasetProcess.ts
```diff
@@ -6,11 +6,12 @@ export default {
     return queryInterface.sequelize.transaction(async t => {
       const tableNames = await queryInterface.showAllTables();
       if (tableNames.indexOf('pipelines') >= 0) {
+        const tables = await queryInterface.describeTable('pipelines');
```
yorkie, line 9: It can be renamed to `table`?

packages/daemon/src/migrations/20200803063219-addDatasetProcess.ts
```diff
@@ -22,8 +23,9 @@ export default {
     return queryInterface.sequelize.transaction(async t => {
       const tableNames = await queryInterface.showAllTables();
       if (tableNames.indexOf('pipelines') >= 0) {
-        await queryInterface.removeColumn('pipelines', 'datasetProcess', { transaction: t });
-        await queryInterface.removeColumn('pipelines', 'datasetProcessParams', { transaction: t });
+        const tables = await queryInterface.describeTable('pipelines');
```
yorkie, line 26: Ditto

packages/daemon/src/migrations/20200803063219-addDatasetProcess.ts
WenheLI: It will get all tables under 'pipelines', `tables` might be the most suitable name

PR #433, opened 2020-08-12
boa: fix crush if import a nonexistent module

FeelyChau:


FeelyChau:
> Add a test case?

Done.

Issue #437, opened 2020-08-12
Actions speed up.


FeelyChau:
This is an actions speed-up scheme.
Using [Artifact](https://docs.github.com/en/actions/configuring-and-managing-workflows/persisting-workflow-data-using-artifacts) and [Cache](https://docs.github.com/en/actions/configuring-and-managing-workflows/caching-dependencies-to-speed-up-workflows) to make actions high-efficiency.

![image](https://user-images.githubusercontent.com/10196611/89995942-c5e06000-dcbc-11ea-87a3-f99c545595dc.png)

yorkie:
What files will be put to artifact?

FeelyChau:
For npm project, [it should be all the workspace](https://github.com/actions/upload-artifact#uploading-to-the-same-artifact).

yorkie:
I think the lint should be in front of the npm-install.

FeelyChau:
> I think the lint should be in front of the npm-install.

Without npm-install, the lint tools will not be installed.

FeelyChau:
CI speed is not a problem anymore.
This issue could be closed.

PR #438, opened 2020-08-12
CI: speed up

FeelyChau:
Remove all plugins and pipboard to [Imgcook](https://github.com/imgcook).
New schema for github actions.
#437


rickycao-qy:
LGTM

PR #446, opened 2020-08-13
CI/CD: release Docker image

liangzr:
This PR will includes:
- Build & Publish a Docker image after each version released.
- Installing a specified version for Pipcook from the tag.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -39,8 +39,7 @@ jobs:
 
           if
             [ $(check_pkg pipcook-cli $ver) == $ver ] &&
-            [ $(check_pkg daemon $ver) == $ver ] &&
-            [ $(check_pkg pipboard $ver) == $ver ]
+            [ $(check_pkg daemon $ver) == $ver ]
           then
             echo ::set-env name=NPM_PUBLISHED::true
           fi
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -38,8 +38,8 @@ jobs:
           }
 
           if
-            [ $(check_pkg pipcook-cli $ver) == $ver ] ||
-            [ $(check_pkg daemon $ver) == $ver ] ||
+            [ $(check_pkg pipcook-cli $ver) == $ver ] &&
+            [ $(check_pkg daemon $ver) == $ver ] &&
             [ $(check_pkg pipboard $ver) == $ver ]
           then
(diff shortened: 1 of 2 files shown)


rickycao-qy:
any versions of cli will install latest versions of daemon and pipboard with current logics. Should we also add version control for pipcook init

liangzr:
> any versions of cli will install latest versions of daemon and pipboard with current logics. Should we also add version control for pipcook init

@rickycao-qy  Agree! A specified version will be very meaningful for Docker to maintain a stable environment.It's will install an uncertain version of pipboard/daemon now depends on the time of the building image.

.github/workflows/release.yml
```diff
@@ -22,7 +22,35 @@ jobs:
...
+          }
+
+          check_image() {
+            curl -sL "https://hub.docker.com/v2/repositories/pipcook/pipcook/tags/$1" | jq -r '.name'
+          }
+
+          if
+            [ $(check_pkg pipcook-cli $ver) == $ver ] &&
+            [ $(check_pkg daemon $ver) == $ver ] &&
+            [ $(check_pkg pipboard $ver) == $ver ]
```
yorkie, line 43: We might not need to check version for pipboard, because we have moved that to a separate repository: https://github.com/imgcook/pipboard.

liangzr: Looks like the release of pipboard has separated with this repo, But if we add version-control for pipcook init in Dockerfile, it will install a specified version for pipboard and maybe failed without verification.

yorkie: I started a discussion at https://github.com/alibaba/pipcook/pull/481/files#r474507514, let's move there :)

FeelyChau:
> > any versions of cli will install latest versions of daemon and pipboard with current logics. Should we also add version control for pipcook init
> 
> @rickycao-qy Agree! A specified version will be very meaningful for Docker to maintain a stable environment.It's will install an uncertain version of pipboard/daemon now depends on the time of the building image.

PR #481 adds the argument to specify the version.

.github/workflows/release.yml
yorkie: As the #481 has been merged which has removed a specific version for the board, IMO we are able to remove board check.

liangzr: Removed.

yorkie (approved):

rickycao-qy (approved):
LGTM

Issue #449, opened 2020-08-13
daemon: migration process takes 27s to get up


yorkie:
```
✔ skiped, daemon is not running.
npx: installed 80 in 27.745s
single process mode is still in experiment, please don't use it in production environment
Server is listening at http://localhost:6927, cost 30.716402275s
```

The `npx` takes too many seconds to install dependencies.

PR #450, opened 2020-08-14
daemon: preinstall sequelize-cli for migration

FeelyChau:
Fix: #449 

### Benchmark
MacOS Catalina 10.15.5 MacBook Pro (16-inch, 2019)
Node: v12.17.0

#### Before:
```shell
$ node ./packages/daemon/bootstrap.js
npx: 80 安装成功，用时 30.633 秒
single process mode is still in experiment, please don't use it in production environment
Server is listening at http://localhost:6927, cost 35.354666849s
```
#### After:
```shell
$ node ./packages/daemon/bootstrap.js
single process mode is still in experiment, please don't use it in production environment
Server is listening at http://localhost:6927, cost 3.240450431s
```

### DIFF
--- a/packages/daemon/package.json
+++ b/packages/daemon/package.json
@@ -20,6 +20,7 @@
     "midway": "^1.19.0",
     "request-promise": "^4.2.5",
     "sequelize": "^5.21.7",
+    "sequelize-cli": "^6.2.0",
     "sqlite3": "^4.1.1",
     "ssestream": "^1.1.0"
   },


yorkie:
We need a benchmark for the bootstrap time of daemon.

FeelyChau:
> We need a benchmark for the bootstrap time of daemon to confirm if #449 gets fixed.

Updated.

FeelyChau:
The installation of sequelize-cli costs too much time (30s), I think the better way is: check if the migration is applied, if not, download sequelize-cli and migrate, else just ignore.

FeelyChau:
> The installation of sequelize-cli costs too much time (30s), I think the better way is: check if the migration is applied, if not, download sequelize-cli and migrate, else just ignore.

At this time, we can't get the Initialized Sequelize object in `bootstrap.js`.

PR #452, opened 2020-08-16
cli: remove duplicated code

shenyiling:


packages/cli/src/bin/pipcook.ts
```diff
@@ -38,7 +38,7 @@ import devPlugin from '../actions/dev-plugin';
     versionStr.push(`Pipcook Daemon  v${daemonPkg.version} ${constants.PIPCOOK_DAEMON_SRC}`);
   }
   if (await pathExists(boardPath)) {
-    const boardPkg = require(join(constants.PIPCOOK_BOARD_SRC, 'package.json'));
+    const boardPkg = require(boardPath);
```
FeelyChau, line 41:
```suggestion
    const boardPkg = await ReadJson(boardPath);
```

shenyiling: what mean of it?why is it different from [require(daemonPath)](https://github.com/alibaba/pipcook/blob/master/packages/cli/src/bin/pipcook.ts#L37)?

FeelyChau: `require` is a sync function, `ReadJson` is async.

FeelyChau:
@shenyiling We should use asynchronous methods whenever possible to ensure concurrency, so the function `require` should be instead with `readJson`.

PR #453, opened 2020-08-16
fix: pipeline install with pyIndex

FeelyChau:


yorkie:
Removed the label `bug` which is used for issue only.

yorkie:
Shall we have tests for this?

FeelyChau:
> Shall we have tests for this?

Maybe we need to collect the logs to confirm the option `pyIndex`.

PR #455, opened 2020-08-17
daemon: pipeline and plugin installation queue

FeelyChau:
At this time, the plugin dir is a nodejs project, we can't update/add/remove the plugins when any pipeline run, as the same, we can't run any pipeline when modify plugins. So, we should queue all the operations of plugins and pipelines.


packages/daemon/src/service/pipeline.ts
```diff
@@ -394,6 +394,17 @@ export class PipelineService {
...
     const plugins = await this.fetchPlugins(pipeline);
-    await this.startJob(job, pipeline, plugins, log);
+    job.status = PipelineStatus.PENDING;
+    await job.save();
+    return new Promise((resolve, reject) => {
+      pluginQueue.push((cb) => {
+        this.startJob(job, pipeline, plugins, log).then(() => {
+          resolve();
+          cb();
+        }).catch((err) => {
```
yorkie, line 404:
```suggestion
        }).catch(reject);
```

packages/daemon/src/service/plugin.ts
```diff
@@ -128,23 +129,31 @@ export class PluginManager {
...
+          this.pluginRT.costa.uninstall(pkg.name);
+          reject(err);
+          cb();
+        });
+      });
+    });
   }
 
-  async installAsync(pkg: PluginPackage, pyIndex?: string, force?: boolean): Promise<TraceResp<PluginResp>> {
+  async installNextTick(pkg: PluginPackage, pyIndex?: string, force?: boolean): Promise<TraceResp<PluginResp>> {
```
yorkie, line 149: Why change the function name?

packages/daemon/src/service/pipeline.ts
```diff
@@ -17,11 +17,11 @@ import {
   PluginStatus
 } from '@pipcook/pipcook-core';
 import { PluginPackage, RunnableResponse, PluginRunnable } from '@pipcook/costa';
-
 import { PipelineModel, PipelineModelStatic } from '../model/pipeline';
 import { JobModelStatic, JobModel } from '../model/job';
 import { PluginManager } from './plugin';
 import { LogObject } from './log-manager';
+import { pluginQueue } from '../utils/queue';
```
yorkie, line 24: Where is the module?

FeelyChau: Fixed.

packages/daemon/src/service/plugin.ts
FeelyChau: Because of the midway context scope, the name `nextTick` is a more specific  description and it's not going to affect the context scope.

yorkie: installAtNextTick?

FeelyChau: `installAtNextTick ` is better.

packages/daemon/src/utils/queue.ts
```diff
@@ -0,0 +1,3 @@
+import Queue from 'queue';
+
+export const pluginQueue = new Queue({ autostart: true, concurrency: 1 });
```
yorkie, line 3: Do we really need this to be a separate file?

FeelyChau: Or merge this into `emitter.ts` and rename to `utils.ts`.

yorkie: I'm +1 on this :)

FeelyChau: Updated.

Issue #445, opened 2020-08-13
Daemon: throw error  `unexpected end of file` when upload a plugin from tarball.


FeelyChau:
Here is the logs:
```shell
Run npm run test:pipeline -- text-bayes-classification
  npm run test:pipeline -- text-bayes-classification
  shell: /bin/bash -e {0}

> pipcook@ test:pipeline /Users/runner/work/pipcook/pipcook
> sh ./run_pipeline.sh "text-bayes-classification"

[start]: run pipeline from ./test/pipelines/text-bayes-classification.json
[start]: installing plugin ./packages/plugins/data-collect/csv-data-collect
[start]: installing @pipcook/plugins-csv-data-collect from /Users/runner/work/pipcook/pipcook/packages/plugins/data-collect/csv-data-collect
[fail]: unexpected end of file
npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! pipcook@ test:pipeline: `sh ./run_pipeline.sh "text-bayes-classification"`
npm ERR! Exit status 1
npm ERR! 
npm ERR! Failed at the pipcook@ test:pipeline script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.

npm ERR! A complete log of this run can be found in:
npm ERR!     /Users/runner/.npm/_logs/2020-08-13T04_52_47_092Z-debug.log
##[error]Process completed with exit code 1.
```
We need more logs to fix this.
#444

yorkie:
Is this fixed by #457?

FeelyChau:
> Is this fixed by #457?

Fixed.

PR #457, opened 2020-08-17
costa: fix upload tarball error

FeelyChau:
Fix bug:
https://github.com/alibaba/pipcook/issues/445
### Test code:
```js
var fs = require('fs-extra');

async function test() {
    const fr = fs.createReadStream('/path/to/read');
    const fw = fs.createWriteStream('/path/to/write');
    fr.pipe(fw);
    fr.on('end', () => {
      console.log('read end');
      resolve();
    });
    fw.on('finish', () => {
      console.log('write finish');
    });
  })
}

test().then(() => {
  console.log('test end');
});
```
### output:
* case A:
```shell
read end
write finish
test end
```
* case B:
```
read end
test end
write finish
```

If we get the case B, the error happens.


yorkie:
Why not add the case in the codebase?

packages/costa/src/runtime.ts
```diff
@@ -292,13 +302,9 @@ export class CostaRuntime {
...
     const filename = path.join(fileDir, 'pkg.tgz');
+    const writeStream = createWriteStream(filename);
     await mkdirp(fileDir);
-    await new Promise((resolve, reject) => {
-      const writeStream = createWriteStream(filename);
-      stream.pipe(writeStream);
-      stream.on('end', resolve);
-      stream.on('error', reject);
-    });
+    await this.saveFile(stream, writeStream);
```
yorkie, line 307: Could you please move the writeStream to saveFile, too?

FeelyChau: If that, the unit test will be unable.

FeelyChau: Moved this function to `utils.ts`.

packages/costa/src/utils.ts
```diff
@@ -17,3 +17,17 @@ export function pipe(readable: NodeJS.ReadableStream, writable: NodeJS.WritableS
     writable.write(data);
   });
 }
+
+/**
+ * pipe read stream to write steam with finish event
+ * @param readStream the read stream
+ * @param writeStream the write stream
+ */
+export function pipeWithFinish(readStream: NodeJS.ReadableStream, writeStream: NodeJS.WritableStream): Promise<void> {
```
yorkie, line 26: The name looks wired a bit, `finish` is just an event.

FeelyChau: How about `pipePromisify`?

packages/costa/src/runtime.ts
```diff
@@ -81,22 +81,20 @@ function extractPackageJsonFromReadable(readable: NodeJS.ReadableStream, pkgFile
...
-    extract.on('error', reject);
     extract.on('entry', (header, stream, next) => {
       if (header.name === pkgFilename) {
         stream.on('data', (buf) => packageJson += buf);
       }
       stream.once('end', next);
       stream.resume();
     });
-    extract.on('finish', () => {
+    pipeWithFinish(readable, extract).then(() => {
```
yorkie, line 91: No necessary to return a Promise here, we could use a callback style and call promisify for async/await use-case.

packages/costa/src/utils.ts
yorkie: I think `safePipe` is more specific.

yorkie: Or `pipeGracefully`?

packages/costa/src/runtime.ts
FeelyChau: This part of the code could also be simplified a little bit.

FeelyChau: Updated.

packages/costa/src/utils.ts
FeelyChau: pipeGracefully +1

PR #458, opened 2020-08-17
add default processor to json dump

rickycao-qy:


packages/boa/lib/index.js
```diff
@@ -80,7 +80,10 @@ function copy(T) {
 function dump(T) {
   return pyInst.import('json')
     .__getattr__('dumps')
-    .invoke(asHandleObject(T));
+    .invoke(asHandleObject(T), {
+      default: _internalWrap(builtins).str,
```
yorkie, line 84: Is this a typo? It prints builtins when dump T?

rickycao-qy: default is the function that will be used when some fields cannot be serialized. By default, we use str method as this method

yorkie: I see, may I ask you to add a comment there to explain?

rickycao-qy: Sure, done

PR #460, opened 2020-08-17
Build: Speed up building

WenheLI:
This PR speeds up building process


packages/cli/src/config.ts
```diff
@@ -7,7 +7,7 @@ export const dependencies: string[] = [
 
 export const isLocal = process.env.NODE_ENV === 'local';
 export const daemonPackage = isLocal ? path.join(__dirname, '../../daemon') : '@pipcook/daemon';
-export const boardPackage = isLocal ? path.join(__dirname, '../../pipboard') : '@pipcook/pipboard';
+export const boardPackage = '@pipcook/pipboard';
```
rickycao-qy, line 10: Should we also delete isLocal mode for daemon above?

yorkie: Yea, we are able to remove local mode absolutely. /cc @FeelyChau

yorkie:
@WenheLI Do we have any data to show how is the speed up?

rickycao-qy:
master: 30.9s
this branch: 3.2s

packages/cli/src/config.ts
FeelyChau: It's done by #472

FeelyChau (approved):
LGTM.

Issue #442, opened 2020-08-12
costa: add retry to install python packages


yorkie:
```
[info]:   Downloading https://files.pythonhosted.org/packages/1a/3b/fa92ece1e58a6a48ec598bab327f39d69808133e5b2fb33002ca754e381e/torch-1.4.0-cp37-cp37m-manylinux1_x86_64.whl (753.4MB)
[warn]: THESE PACKAGES DO NOT MATCH THE HASHES FROM THE REQUIREMENTS FILE. If you have updated the package versions, please update the hashes. Otherwise, examine the package contents carefully; someone may have tampered with them.
[warn]:     torch==1.4.0 from https://files.pythonhosted.org/packages/1a/3b/fa92ece1e58a6a48ec598bab327f39d69808133e5b2fb33002ca754e381e/torch-1.4.0-cp37-cp37m-manylinux1_x86_64.whl#sha256=8fff03bf7b474c16e4b50da65ea14200cc64553b67b9b2307f9dc7e8c69b9d28:
[warn]:         Expected sha256 8fff03bf7b474c16e4b50da65ea14200cc64553b67b9b2307f9dc7e8c69b9d28
[warn]:              Got        aa5155c34a27781e23b5ce784f027076a602b2d7b1c72eda3e131dfd15157017
[warn]: 
[warn]: You are using pip version 19.0.3, however version 20.2.2 is available.
[warn]: You should consider upgrading via the 'pip install --upgrade pip' command.
[fail]: invalid code 1 from /home/runner/.pipcook/plugins/conda_envs/@pipcook/plugins-detectron-fasterrcnn-model-define@1.1.0/bin/pip3
```

If we were to support retry after failure, we could decrease the failed times.

FeelyChau:
Python packages are not installed by costa at Pipcook v2.
They are packaged in the framework now.

PR #462, opened 2020-08-17
ci: retry when failed

FeelyChau:


yorkie:
Retries on CI is not enough that in our use case, we should confirm this mechanism is also applied to install plugins specifically.

FeelyChau:
> Retries on CI is not enough that in our use case, we should confirm this mechanism is also applied to install plugins specifically.

Yes, this is what the issue #442 wants to do, before the installation-retry mechanism implement, we could add the action-retry-plugin to avoid too much human intervention because of network.

yorkie:
No, #442 is not to add retry to CI and not fix the CI problems :(

FeelyChau:
> No, #442 is not to add retry to CI and not fix the CI problems :(

What I mean `this` is what you want to do.
```
we should confirm this mechanism is also applied to install plugins specifically.
```

PR #465, opened 2020-08-18
daemon: add benchmark for startup

FeelyChau:
We can make daemon startup benchmark by `npm run benchmark`.


yorkie:
I think "packages/daemon/benchmark/bootstrap.js" is a better choice.

FeelyChau:
> I think "packages/daemon/benchmark/bootstrap.js" is a better choice.

Renamed.

PR #468, opened 2020-08-18
fix data process to predict

rickycao-qy:
Since we have changed how to data process works, need to change predict script accordingly

### DIFF
--- a/packages/daemon/templates/node/predict.js
+++ b/packages/daemon/templates/node/predict.js
@@ -29,9 +29,9 @@ async function predict(data) {
   if (!model) {
     model = await modelDefine;
   }
-  const sample = { data, label: null };
+  let sample = { data, label: null };
   if (typeof dataProcessModule === 'function') {
-    await dataProcessModule(sample, {}, JSON.parse(pipeline.dataProcessParams));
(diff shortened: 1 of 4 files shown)


packages/daemon/templates/node/predict.js
```diff
@@ -29,9 +29,9 @@ async function predict(data) {
   if (!model) {
     model = await modelDefine;
   }
-  const sample = { data, label: null };
+  let sample = { data, label: null };
   if (typeof dataProcessModule === 'function') {
-    await dataProcessModule(sample, {}, JSON.parse(pipeline.dataProcessParams));
+    sample = await dataProcessModule(sample, {}, JSON.parse(pipeline.dataProcessParams));
```
yorkie, line 34: Dos the __data-process__ plugin return `Sample` now?

rickycao-qy: Exactly, data-process and global-data-process will not modify original data.

yorkie: See https://github.com/alibaba/pipcook/blob/master/docs/spec/plugin.md#plugin-category, this needs an update IMO.

rickycao-qy: Done ~~

yorkie (approved):

PR #469, opened 2020-08-18
boa: fix memory leak in js heap

yorkie:
This removes the reference for every `boa.Object`, it reduces the memory consumption when creating Python object repeatedly.

Screen Shot 2020-08-18 at 9 13 49 PM

Without this change, the _mlstep1/2_ will have 2000+ `boa.Object`, because every object will be referenced to global handles/gc roots, that's one of the reasons for memory leaks.


yorkie:
There is a weird problem on our Linux build, I have tried using only 1 script with tap command, it seems still producing core dump, but running with node directly is always working.

I have also opened an issue at https://github.com/tapjs/node-tap/issues/693, we need to continue to address the problem.

yorkie:
I have addressed the current Linux build issue is not related to this patch, see https://github.com/tapjs/node-tap/issues/693, the official examples(napi-only, node-addon-api) for wrapping c++ object would generate the same core dump, so we have the following plan:

1. change our test framework to `tape` instead of `tap`.
2. wait for `tap` to fix this issue.

Any thoughts? @FeelyChau @rickycao-qy

FeelyChau:
> * change our test framework to `tape` instead of `tap`.

```
change our test framework to tape instead of tap.
```
+1

yorkie:
I have migrated from tap to tape, and disable the esm testing for Node12, which causes the same error on Node12, but working on later versions.

rickycao-qy:
> I have addressed the current Linux build issue is not related to this patch, see [tapjs/node-tap#693 (comment)](https://github.com/tapjs/node-tap/issues/693#issuecomment-677317775), the official examples(napi-only, node-addon-api) for wrapping c++ object would generate the same core dump, so we have the following plan:
> 
> 1. change our test framework to `tape` instead of `tap`.
> 2. wait for `tap` to fix this issue.
> 
> Any thoughts? @FeelyChau @rickycao-qy

Sounds good

PR #470, opened 2020-08-18
ci: set global timeout

FeelyChau:


run_tests.ts
```diff
@@ -12,9 +12,11 @@ const indexSpec = process.argv.indexOf('--spec');
 if (indexSpec !== -1 && process.argv[indexSpec + 1]) {
   spec = [ process.argv[indexSpec + 1] ];
 }
+
 const runner = new JasmineCtor();
+runner.jasmine.DEFAULT_TIMEOUT_INTERVAL = 20 * 6000;
```
yorkie, line 17: 20 seconds?

run_tests.ts
```diff
@@ -12,9 +12,11 @@ const indexSpec = process.argv.indexOf('--spec');
...
 runner.loadConfig({
-  spec_files: spec || [ 'src/**/*_test.ts', 'test/**/*.ts' ],
+  spec_files: spec || [ 'src/**/*_test.ts', 'test/**/job.ts' ],
```
yorkie, line 19: Why using the job instead of *?

FeelyChau: It's test code, need to be removed.

run_tests.ts
FeelyChau: Modified to 20 mins.

FeelyChau:
@yorkie  Ready.

PR #471, opened 2020-08-19
fix data batch

rickycao-qy:
There is bug in dataloader:
  the fetchIndex (pointer to current data) is not updated


yorkie (dismissed):
LGTM

yorkie:
And removed the label `bug` which is for issue-only.

FeelyChau:
> And removed the label `bug` which is for issue-only.

Maybe we should create a new label `bug-fix`.

yorkie:
> > And removed the label `bug` which is for issue-only.
> 
> Maybe we should create a new label `bug-fix`.

Why do we need this label?

WenheLI (approved):
LGTM

PR #476, opened 2020-08-19
cli: fix plugin install error, fix ci test case

FeelyChau:
The command `pipcook pipeline install /path/to/pipeline.json` finish with error, and CI did not check out this, because the plugins in the case are all installed before.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -72,7 +72,7 @@ jobs:
         npm run test:pipeline -- text-bayes-classification
     - name: install plugins from specific pipeline
       run: |
-          ./packages/cli/dist/bin/pipcook pipeline install ./example/pipelines/text-bayes-classification.json
+          ./packages/cli/dist/bin/pipcook pipeline install ./example/pipelines/chinese-poem-creation.json
           ./packages/cli/dist/bin/pipcook pipeline install https://raw.githubusercontent.com/alibaba/pipcook/master/example/pipelines/text-bayes-classification.json
     - name: pipcook run from local filename
       run: |
--- a/packages/cli/src/service/pipeline.ts
+++ b/packages/cli/src/service/pipeline.ts
@@ -111,12 +111,12 @@ export async function install(filename: string, opts: any): Promise<void> {
     // check the installation
     logger.info('check plugin:');
     let isSuccess = true;
-    for (const plugin of constants.PLUGINS) {
+    await Promise.all(constants.PLUGINS.map(async (plugin) => {
       // if use pipeline[plugin], error throw:
       // TS2536: Type 'PluginTypeI' cannot be used to index type 'PipelineResp'.
       const pluginName = (pipeline as any)[plugin];
       if (!pluginName) {
-        continue;
+        return;
       }
       const plugins = await client.plugin.list({ name: pluginName });
       if (plugins.length > 0 && plugins[0].status === PluginStatus.INSTALLED) {
@@ -125,7 +125,7 @@ export async function install(filename: string, opts: any): Promise<void> {
         isSuccess = false;
         logger.fail(`${pluginName} ${PluginStatusValue[plugins[0]?.status || PluginStatus.FAILED]}.`, false);
       }
-    }
+    }));
     if (isSuccess) {
       logger.success('pipeline installed successfully.');
     } else {
--- a/packages/cli/src/service/pipeline.ts
+++ b/packages/cli/src/service/pipeline.ts
@@ -72,10 +72,8 @@ export async function installPackageFromConfig(config: any, opts: any): Promise<
   for (const plugin of constants.PLUGINS) {
     const packageName = config.plugins[plugin]?.package;
     if (typeof packageName === 'string') {
-      if (packageName[0] === '.') {
-        const pkg = await pluginInstall(packageName, opts);
-        config.plugins[plugin].package = pkg.name;
-      }
+      const pkg = await pluginInstall(packageName, opts);
+      config.plugins[plugin].package = pkg.name;
     }
   }
 }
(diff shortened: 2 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -72,7 +72,7 @@ jobs:
         npm run test:pipeline -- text-bayes-classification
     - name: install plugins from specific pipeline
       run: |
-          ./packages/cli/dist/bin/pipcook pipeline install ./example/pipelines/text-bayes-classification.json
+          ./packages/cli/dist/bin/pipcook pipeline install ./example/pipelines/chinese-poem-creation.json
```
yorkie, line 75: What's the reason to change the test pipeline?

yorkie:
> The command `pipcook pipeline install /path/to/pipeline.json` finish with error, and CI did not check out this, because the plugins in the case are all installed before.

May I ask you to point out where the current pipeline not checking?

packages/cli/src/service/pipeline.ts
```diff
@@ -72,10 +72,8 @@ export async function installPackageFromConfig(config: any, opts: any): Promise<
   for (const plugin of constants.PLUGINS) {
     const packageName = config.plugins[plugin]?.package;
     if (typeof packageName === 'string') {
-      if (packageName[0] === '.') {
-        const pkg = await pluginInstall(packageName, opts);
-        config.plugins[plugin].package = pkg.name;
-      }
+      const pkg = await pluginInstall(packageName, opts);
```
yorkie, line 75: This might be better to put in another PR :)

packages/cli/src/service/pipeline.ts
```diff
@@ -113,12 +111,12 @@ export async function install(filename: string, opts: any): Promise<void> {
     // check the installation
     logger.info('check plugin:');
     let isSuccess = true;
-    await constants.PLUGINS.map(async (plugin) => {
+    for (const plugin of constants.PLUGINS) {
```
yorkie, line 114: This will remove the concurrent running, do we have any specific reason for that?

FeelyChau: The plugin project can't run concurrently, and the `await` can't take effect.

FeelyChau:
> > The command `pipcook pipeline install /path/to/pipeline.json` finish with error, and CI did not check out this, because the plugins in the case are all installed before.
> 
> May I ask you to point out where the current pipeline not checking?

```shell
[start]: start install pipeline from ./example/pipelines/text-bayes-classification.json
[start]: installing plugins
[info]: start to create pipeline
[success]: pipeline is created: x8xo318f
[info]: check plugin:
[success]: pipeline installed successfully.
[success]: @pipcook/plugins-csv-data-collect installed.
[success]: @pipcook/plugins-csv-data-access installed.
[success]: @pipcook/plugins-bayesian-model-define installed.
[success]: @pipcook/plugins-bayesian-model-train installed.
[success]: @pipcook/plugins-bayesian-model-evaluate installed.
```
The log should be:
```shell
[success]: @pipcook/plugins-csv-data-collect installed.
[success]: @pipcook/plugins-csv-data-access installed.
[success]: @pipcook/plugins-bayesian-model-define installed.
[success]: @pipcook/plugins-bayesian-model-train installed.
[success]: @pipcook/plugins-bayesian-model-evaluate installed.
[success]: pipeline installed successfully.
```
If the pipeline plugins have not been install, it will throw error here.

packages/cli/src/service/pipeline.ts
FeelyChau: The original intent here was to install the plugin serial.

packages/cli/src/service/pipeline.ts
FeelyChau: This fixes an error that would not have been installed if the  name was not a local path.

packages/cli/src/service/pipeline.ts
yorkie: Questions: a) What do you mean by the "plugin project" b) why does the `await` not taking effect?

.github/workflows/build.yml
yorkie, line 75: So why do we change the pipeline to test here?

yorkie:
@FeelyChau We have a label named `tests`, maybe that's more specific than `ci/cd` which is related to our workflows.

.github/workflows/build.yml
FeelyChau: This makes a fresh installation with no plugins ware installed.

FeelyChau:
> @FeelyChau We have a label named `tests`, maybe that's more specific than `ci/cd` which is related to our workflows.

OK, changed.

packages/cli/src/service/pipeline.ts
yorkie: What error on that case?

packages/cli/src/service/pipeline.ts
FeelyChau: a) "plugin project" means the plugin nodejs project, we can't install package concurrently by mutil-process.
b) we can't await a map funtion.
test code:
```js
const arr = [1, 2, 3];
async function test() {
  await arr.map(async n => {
    setTimeout(() => console.log(n), 100);
  })
  console.log(4);
}

test();

```
output:
```shell
4
1
2
3
```

FeelyChau: Question A may be not a problem after #455 , I'll test it.

packages/cli/src/service/pipeline.ts
FeelyChau: `installPackageFromConfig ` should install plugins whatever the plugin name protocol(local path, url, git or npm package name).
But it accepts local path only.

yorkie: Ok, resolved.

yorkie (dismissed):

FeelyChau:
@yorkie The parallel installation has been restored. May I ask you to review again?

yorkie (approved):
LGTM

PR #477, opened 2020-08-19
costa: install Node and Python packages in parallel

FeelyChau:
Benchmark:
Test command:
```shell
$ time ./packages/cli/dist/bin/pipcook plugin install @pipcook/plugins-tensorflow-resnet-model-define --tuna
```

This plugin depends on `tensorflow@2.2.0`.

| Branch                       | Test A          | Test B          | Test C          | Test D          | AVG             |
| ---------------------------- | --------------- | --------------- | --------------- | --------------- | --------------- |
| master                       | 1 m 20.06 s | 1 m 19.86 s | 1 m 25.03 s | 1 m 18.33 s | 1 m 20.82 s |
| modify/parallel-installation | 1 m 11.06 s | 1 m 04.47 s | 1 m 04.92 s | 1 m 11.27 s | 1 m 07.93 s |

In this typical scenario, reduced installation time by 15%.

### DIFF
--- a/packages/costa/benchmark/makefile
+++ b/packages/costa/benchmark/makefile
@@ -2,3 +2,5 @@ export DEBUG=costa*
 
 bootstrap:
 	time node ./bootstrap.js
+install:
+	time node ./plugin-install.js
--- a/packages/costa/benchmark/plugin-install.js
+++ b/packages/costa/benchmark/plugin-install.js
@@ -11,7 +11,6 @@ const costa = new CostaRuntime({
 });
 
 (async () => {
-  const start = Date.now();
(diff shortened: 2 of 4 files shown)


yorkie (approved):
LGTM and the results are reasonable because this PR merges the cost of installing node_modules into Python's, the former takes 10-20s approximately to be done :)

yorkie:
That would be that great if we add this benchmark script in this PR :)

FeelyChau:
> That would be that great if we add this benchmark script in this PR :)

Done.

yorkie (approved):

Issue #348, opened 2020-07-15
boa: support the esm for node.js >=14.5.0


yorkie:
https://github.com/nodejs/node/pull/33501 removed the ESM dynamic module format, and I skip the esm tests at https://github.com/alibaba/pipcook/pull/347, but we still need to add support for node(>=14.5.0), just ping to us if you are interested in this.

NHibiki:
@yorkie I am interested in this.

PR #479, opened 2020-08-20
feat: support esm loader

NHibiki:
Related to #348 

This PR enables hybrid exports of boa module which allows both `commonjs` and `esm` import.

Furthermore, test cases are also added to test the import of the boa main entry.


yorkie:
@NHibiki To fix #348, the correct way is to use `getSource` and `getGlobalPreloadCode` instead of the deprecated dynamic format at our [custom loader](https://github.com/alibaba/pipcook/blob/master/packages/boa/esm/loader.mjs#L18), @rickyes may I ask you to have guidance?

rickyes:
@NHibiki 
- Maybe you can first understand how Pipcook's custom `esm loader` works, refer to article https://cnodejs.org/topic/5ec4b241a87fc8583363d488. 
- The problem is that `dynamicInstantiate` hook have been deprecated in Node.js v14.8.0, so we need a new way to support custom `esm loader`, and we need to be compatible with < Node.js v14.8.0.

Please feel free to ask me if you have any problems :)

yorkie:
https://github.com/alibaba/pipcook/blob/master/packages/boa/tests/es-module-loaders/common.js#L23-L29 This is able to be removed.

NHibiki:
> https://github.com/alibaba/pipcook/blob/master/packages/boa/tests/es-module-loaders/common.js#L23-L29 This is able to be removed.

OK, let me remove it and run a test on it.

yorkie:
@NHibiki @rickyes Is this way working with <=14?

packages/boa/esm/loader.mjs
```diff
@@ -37,3 +43,25 @@ export function dynamicInstantiate(url) {
...
+    return {
+      source: `
+import boa from '../lib/index.js';
+
+const moduleInstance = boa.import('${url.replace(protocol, '')}');
+
+${moduleExports.map(name => `const __boa__${name} = moduleInstance['${name}'];`).join('\n')}
+
+export {
+${moduleExports.map(name => `  __boa__${name} as ${name},`).join('\n')}
```
yorkie, line 61:
```suggestion
${moduleExports.map(name => `  ${name} = moduleInstance['${name}'],`).join('\n')}
```

packages/boa/esm/loader.mjs
```diff
@@ -37,3 +43,25 @@ export function dynamicInstantiate(url) {
...
+};
+export default moduleInstance;`
+    };
+  }
+  return defaultGetSource(url, context, defaultGetSource);
+}
```
yorkie, line 67: EOF

packages/boa/esm/loader.mjs
```diff
@@ -15,14 +19,16 @@ export function getFormat(url, context, defaultGetFormat) {
...
+      format: 'module'
     }
   }
 
   // Other protocol are assigned to nodejs for internal judgment loading
   return defaultGetFormat(url, context, defaultGetFormat);
 }
 
+// deprecated by node v14.8.0
 export function dynamicInstantiate(url) {
```
yorkie, line 32: Could this be removed, too?

NHibiki:
> @NHibiki @rickyes Is this way working with <=14?

I tested this method in node v12 and it works

packages/boa/esm/loader.mjs
NHibiki: I think export might follow the syntax as:

```js
export {
  foo1 as foo2
}
```

I export in this way because:

- `foo1` should be a declared constant/variable.
- `__boa__` prefix is to prevent re-declaring built-in namespaces in javascript (e.g. `import`, `var`, `eval` which are likely to be used in python)

packages/boa/esm/loader.mjs
NHibiki: Sure, removed 👍

packages/boa/esm/loader.mjs
yorkie: Resolved.

yorkie (dismissed):
Looks good to me if CI is green :)

yorkie:
@rickyes May I ask you to have a review again?

PR #481, opened 2020-08-21
cli: init with version

FeelyChau:
Use `pipcook init 1.1.0` to install daemon and pipboard of v1.1.0.

### DIFF
--- a/packages/cli/src/actions/init.ts
+++ b/packages/cli/src/actions/init.ts
@@ -72,7 +72,7 @@ async function initPlugin(daemonDir: string, pluginDir: string) {
 /**
  * install all dependencies of pipcook into working dir
  */
-const init: InitCommandHandler = async ({ client, tuna, V: version }) => {
+const init: InitCommandHandler = async ({ beta, client, tuna, V: version }) => {
   let npmClient = 'npm';
   if (beta) {
     version = 'beta';
--- a/packages/cli/src/actions/init.ts
+++ b/packages/cli/src/actions/init.ts
@@ -135,7 +135,6 @@ const init: InitCommandHandler = async (version: string, { beta, client, tuna })
     let board = boardPackage;
     if (version) {
       daemon += `@${version}`;
-      board += `@${version}`;
     }
     await Promise.all([
       npmInstall(npmClient, daemon, CoreConstants.PIPCOOK_DAEMON, npmInstallEnvs),
--- a/packages/cli/src/actions/init.ts
+++ b/packages/cli/src/actions/init.ts
@@ -14,11 +14,8 @@ const {
   BOA_CONDA_MIRROR
 } = Constants;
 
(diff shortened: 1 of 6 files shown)


packages/cli/src/bin/pipcook.ts
```diff
@@ -46,7 +46,7 @@ import devPlugin from '../actions/dev-plugin';
   program
     .command('init')
     .option('-c, --client <string>', 'specify your npm client.')
-    .option('-b, --beta', 'use or update the beta version')
```
yorkie, line 49: We can't deprecate the option directly, it's good to bypass it the version.

yorkie:
How about this way to avoid the --version conflicts:

- `pipcook init`
- `pipcook init beta` or `pipcook init --beta`
- `pipcook init 1.2.0`

FeelyChau:
> How about this way to avoid the --version conflicts:
> 
> * `pipcook init`
> * `pipcook init beta` or `pipcook init --beta`
> * `pipcook init 1.2.0`

`pipcook init beta`/`pipcook init 1.2.0` conflicts to `pipcook init --beta`/`pipcook init`.
The command argument can not be null to commander.

yorkie:
See https://github.com/tj/commander.js#commands, we can define just like `command('init [version]')` then the `version` is optional.

packages/cli/src/bin/pipcook.ts
FeelyChau: Updated.

FeelyChau:
> See https://github.com/tj/commander.js#commands, we can define just like `command('init [version]')` then the `version` is optional.

OK, updated.

yorkie:
The CLI Documentation also requires update :)

packages/cli/src/actions/init.ts
```diff
@@ -125,10 +131,15 @@ const init: InitCommandHandler = async ({ client, beta, tuna }) => {
         spaces: 2
       });
     }
-
+    let daemon = daemonPackage;
+    let board = boardPackage;
+    if (version) {
+      daemon += `@${version}`;
+      board += `@${version}`;
```
yorkie, line 138: As @liangzr mentioned at #446, using the same version for daemon and board would cause that we need to bind the board & daemon release. I have no idea about this until we publish the board online.

yorkie: How about split the board version from this PR?

FeelyChau: Do you mean the init action don't install the pipboard?

liangzr: It depends on the rule of dependence between their versions, maybe we should add a `peerDependencies` in `daemon/package.json`, and check it in CI (as a precondition for Docker building)

yorkie:
@liangzr I have removed the version slot to initialize board, and later the board will be on Web so that we don't need to install and build board on `pipcook init`.

yorkie (approved):

liangzr (approved):
LGTM : )

Issue #493, opened 2020-08-24
daemon: failed to start daemon


yorkie:
I just got the following command after install the daemon.

```
$ npx sequelize-cli db:migrate

Sequelize CLI [Node: 13.14.0, CLI: 6.2.0, ORM: 5.22.3]

Loaded configuration file "dist/config/config.js".
== 20200803063219-addDatasetProcess.d: migrating =======

ERROR: Cannot use import statement outside a module
```

yorkie:
This is fixed by #494.

PR #494, opened 2020-08-24
daemon: avoid generating d.ts files

yorkie:
Daemon is an executable so that the d.ts is unnecessary, this PR sets declaration to `false` to avoid this.


yorkie:
This is also to fix #493.

WenheLI (approved):
LGTM

FeelyChau (approved):
LGTM.

PR #495, opened 2020-08-24
daemon: enable cors for ng-pipboard

yorkie:
Our next-generation board will be an absolute web-based application, so we need a CORS setting on daemon for that access to the local/remote Pipcook via SDK.


FeelyChau (approved):
LGTM.

PR #500, opened 2020-08-24
daemon, sdk: add new func to fetch plugin

yorkie:


FeelyChau:
The plugin controller test cases are working on #488 , could you just add the cases for additional api?

yorkie:
> #488

I can't do that because installing is required at first, I think we could rebase them at #488.

FeelyChau (approved):
LGTM.

PR #504, opened 2020-08-25
sdk: fix pipeline update params

FeelyChau:


yorkie:
Why not change the test case's config?

yorkie:
And is this a breaking change?

FeelyChau:
> Why not change the test case's config?

Updated.

FeelyChau:
> And is this a breaking change?

The parameter config was used as type `PipelineConfig`, so this PR will not change the user cases.

yorkie (approved):

PR #508, opened 2020-08-25
sdk, daemon: add fetchByName api and tests

yorkie:


packages/sdk/src/plugin.ts
```diff
@@ -35,6 +35,14 @@ export class Plugin extends BaseApi {
     return get(`${this.route}/${id}/metadata`);
   }
 
+  /**
+   * fetch specific plugin metadata by name
+   * @param name the plugin name, for example "@pipcook/plugins-image-classification-data-collect"
+   */
+  fetchByName(name: string): Promise<any> {
+    return get(`${this.route}/metadata?name=${name}`);
```
FeelyChau, line 43:
```suggestion
    return get(`${this.route}/metadata`, { name });
```

FeelyChau (changes requested):

packages/daemon/test/controller/plugin.test.ts
```diff
@@ -57,4 +57,23 @@ describe('test plugin controller', () => {
       .get(`/api/plugin/not-exists/metadata`)
       .expect(404);
   });
+
+  it('should fetch by name', () => {
+    return app
+      .httpRequest()
+      .get(`/api/plugin/metadata?name=${name}`)
```
FeelyChau, line 64: This should be updated too.

FeelyChau (changes requested):

FeelyChau (approved):

PR #510, opened 2020-08-25
daemon, cli: fix get pipeline config

FeelyChau:


packages/daemon/src/app/controller/pipeline.ts
```diff
@@ -106,6 +106,19 @@ export class PipelineController extends BaseEventController {
     this.ctx.success(json);
   }
 
+  /**
+   * find a pipeline by id
+   */
+  @get('/:id')
+  public async get() {
```
yorkie, line 113: May I ask you not to change the `get()` behavior?

yorkie (changes requested):
And the test case is required, too.

packages/sdk/src/pipeline.ts
```diff
@@ -18,13 +18,22 @@ export class Pipeline extends BaseApi {
...
-    return get(`${this.route}/${id}`);
+    return get(`${this.route}/${id}/config`);
   }
   info = this.get;
+
+  /**
+   * get pipeline info by pipeline id
+   * @param id pipeline id
+   */
+  getPipeline(id: string): Promise<PipelineConfig> {
```
yorkie, line 33:
```suggestion
  getConfig(id: string): Promise<PipelineConfig> {
```

yorkie:
Controller tests should be added, too :)

run_tests.sh
```diff
@@ -2,4 +2,6 @@ lerna run test \
   --scope @pipcook/pipcook-core \
   --scope @pipcook/boa \
   --scope @pipcook/costa \
-  --scope @pipcook/daemon
+  --scope @pipcook/daemon \
+  --scope @pipcook/sdk
```
yorkie, line 6: This should be removed from this PR

yorkie (dismissed):
LGTM if CI is green, and still marked the label "breaking-change" though we don't apply it on sdk and daemon api.

packages/daemon/test/controller/pipeline.test.ts
```diff
@@ -1,4 +1,7 @@
 import { app } from 'midway-mock/bootstrap';
+import { PipelineService } from '../../src/service/pipeline';
+import * as sinon from 'sinon';
+import assert from 'midway-mock/node_modules/@types/power-assert';
```
yorkie, line 4: This line should be removed and import `assert` from `midway-mock/bootstrap`.

yorkie: I have fixed at https://github.com/alibaba/pipcook/pull/510/commits/0b630306e3932ead1f0ef39313791de1118bdb26.

yorkie (approved):

PR #511, opened 2020-08-26
core: add util plugin name parser

FeelyChau:


packages/core/src/types/plugins.ts
```diff
@@ -10,6 +10,12 @@ interface InsertParams {
 
 export type PluginTypeI = 'dataCollect' | 'dataAccess' | 'dataProcess' | 'datasetProcess' | 'modelLoad' | 'modelDefine' | 'modelTrain' | 'modelEvaluate';
 
+// TODO(feely): remove 'fs'
```
yorkie, line 13: Why do we remove the fs support?

packages/core/src/utils/public.ts
```diff
@@ -321,3 +322,23 @@ export function shuffle(array: any[]): void {
...
+  if (path.isAbsolute(name)) {
+    protocol = 'fs';
+  } else if (/^git(\+ssh|\+https|\+http)?:$/.test(urlObject.protocol)) {
+    protocol = 'git';
+  } else if ([ 'https:', 'http:' ].indexOf(urlObject.protocol) !== -1) {
+    protocol = 'tarball';
+  } else if (name[0] !== '.') {
+    protocol = 'npm';
+  } else {
+    protocol = undefined;
```
yorkie, line 341: Throw an error?

packages/core/src/utils/public.ts
```diff
@@ -321,3 +322,23 @@ export function shuffle(array: any[]): void {
...
+ * parse plugin name, return the plugin protocol and url object
+ */
+export function parsePluginName(name: string): { protocol: PluginProtocol; urlObject: url.UrlWithStringQuery } {
+  const urlObject = url.parse(name);
+  let protocol: PluginProtocol;
+  if (path.isAbsolute(name)) {
+    protocol = 'fs';
+  } else if (/^git(\+ssh|\+https|\+http)?:$/.test(urlObject.protocol)) {
+    protocol = 'git';
+  } else if ([ 'https:', 'http:' ].indexOf(urlObject.protocol) !== -1) {
```
yorkie, line 336: This doesn't mean tarbar, just for http/https url.

packages/core/src/utils/public.ts
```diff
@@ -321,3 +322,23 @@ export function shuffle(array: any[]): void {
 export function generateId(): string {
   return nanoid();
 }
+
+/**
+ * parse plugin name, return the plugin protocol and url object
+ */
+export function parsePluginName(name: string): { protocol: PluginProtocol; urlObject: url.UrlWithStringQuery } {
```
yorkie, line 329: public_utils is mostly for the plugin developer, so IMO this function should be another utils module.

packages/costa/src/runtime.ts
```diff
@@ -247,22 +246,22 @@ export class CostaRuntime {
   async fetch(name: string): Promise<PluginPackage> {
     let pkg: PluginPackage;
     const source = this.getSource(name);
-    if (source.from === 'npm') {
+    if (source.protocol === 'npm') {
```
yorkie, line 249: May I ask you not to change the field name?

packages/core/src/utils/public.ts
FeelyChau: For plugin, it's tarball.

packages/core/src/types/plugins.ts
FeelyChau: OK, the `fs` protocol is for tarball-stream which upload from client, rename to `tarball-stream`?

packages/costa/src/runtime.ts
FeelyChau: I think the field name is not very accurate. IMO, `protocol` + `uri` equal `from`, `protocol` is a little bit more accurate.

packages/core/src/utils/public.ts
FeelyChau: OK, should we create a new utils package?

packages/core/src/types/plugins.ts
FeelyChau: Protocol is the concept that needs to be passed to the user,`fs` is the implementation, should it be changed to a more understandable name, like `tarball-stream` or `local-tarball`?

packages/core/src/utils/public.ts
FeelyChau: Updated.

packages/core/src/utils/public.ts
yorkie: I think we need to verify if it ends with .tgz or .gz, just checking for protocol is not enough in this case.

packages/core/src/types/plugins.ts
yorkie: Ok, you should update what you commented before in the TODO.

packages/core/src/utils/public.ts
yorkie: No new package, they could be put in core/src/utils.

packages/costa/src/runtime.ts
yorkie: This PR is not to change it, and it makes some unexpected changes.

packages/core/src/utils/public.ts
FeelyChau: OK.

packages/core/src/utils/public.ts
FeelyChau: OK, it's better to check the path.

packages/core/src/types/plugins.ts
```diff
@@ -10,6 +10,11 @@ interface InsertParams {
 
 export type PluginTypeI = 'dataCollect' | 'dataAccess' | 'dataProcess' | 'datasetProcess' | 'modelLoad' | 'modelDefine' | 'modelTrain' | 'modelEvaluate';
 
+/**
+ * The plugin installation protocol
+ */
+export type PluginProtocol = 'fs' | 'npm' | 'git' | 'tarball';
```
yorkie, line 16:
```suggestion
export type PluginSourceProtocol = 'fs' | 'npm' | 'git' | 'tarball';
```

packages/core/src/index.ts
```diff
@@ -60,6 +61,10 @@ export {
   generateId
 } from './utils/public';
 
+export {
+  parsePluginName
+} from './utils/utils';
```
yorkie, line 66: Hmm, `utils` is duplicated with the folder, suggestion: `internals` or `common`.

packages/costa/src/runtime.ts
```diff
@@ -477,32 +476,24 @@ export class CostaRuntime {
...
+    } else if (protocol === 'npm') {
       src.schema = this.getNameSchema(name);
-      src.from = 'npm';
       let { npmRegistryPrefix } = this.options;
       if (npmRegistryPrefix.slice(-1) === '/') {
         npmRegistryPrefix = npmRegistryPrefix.slice(0, -1);
       }
       src.uri = `${npmRegistryPrefix}/${src.schema.packageName}`;
-    } else {
-      throw new TypeError(`Unsupported resolving plugin name: ${name}`);
```
yorkie, line 505: I think we still need the throw.

yorkie:
LGTM except for the filename, @FeelyChau could you please rename it to a proper filename?

FeelyChau:
> LGTM except for the filename, @FeelyChau could you please rename it to a proper filename?

`common.ts`?
Or change the tree to:
```shell
src
|--public.ts
|--utils.ts
```

yorkie (approved):
LGTM if CI is green

PR #512, opened 2020-08-26
fix python runtime

rickycao-qy:
When the plugin specifies 'python' runtime, boa will try to import plugin's entry python file. 

Right now it imports like:

```
const fn = boa.import(`node_modules.${pkg.name}`);
```

However, if the plugin name includes scope, for instance:

```
@pipcook/ner_model
```

the contents inside import will be

```
node_modules.@pipcook/ner_model
```

The right way to import should be:

```
node_modules.@pipcook.ner_model
```

This PR is to fix this

### DIFF
--- a/packages/costa/src/client/loaders/python.ts
+++ b/packages/costa/src/client/loaders/python.ts
@@ -3,7 +3,7 @@ const boa = require('@pipcook/boa');
 const sys = boa.import('sys');
 
 export default function (pkg: PluginPackage): Function {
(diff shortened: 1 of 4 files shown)


yorkie:
@rickycao-qy May I ask you to add a new fixture to cover this change?

rickycao-qy:
> @rickycao-qy May I ask you to add a new fixture [here](https://github.com/alibaba/pipcook/tree/master/packages/costa/test/plugins) to cover this change?

done

yorkie (approved):

PR #515, opened 2020-08-26
daemon: create plugin record when create pipeline

FeelyChau:
Migration test added: https://github.com/alibaba/pipcook/blob/daemon/create-pipeline/.github/workflows/build.yml#L63


packages/sdk/src/utils.ts
```diff
@@ -1,4 +1,4 @@
 // FIXME(feely): move these to core
 export const tunaMirrorURI = 'https://pypi.tuna.tsinghua.edu.cn/simple';
 export const JobStatusValue = [ 'creating', 'running', 'success', 'fail', 'canceled' ];
-export const PluginStatusValue = [ 'installing', 'installed', 'failed', 'pending' ];
+export const PluginStatusValue = [ 'installing', 'installed', 'failed', 'pending', 'init' ];
```
yorkie, line 4:
```suggestion
export const PluginStatusValue = [ 'installing', 'installed', 'failed', 'pending', 'initialized' ];
```

yorkie:
How can we verify the migration scripts are working correctly?

FeelyChau:
> How can we verify the migration scripts are working correctly?

Migration CI is needed.

yorkie:
> > How can we verify the migration scripts are working correctly?
> 
> Migration CI is needed.

Shall we add that check in this PR?

packages/daemon/src/app/controller/pipeline.ts
```diff
@@ -40,8 +40,26 @@ export class PipelineController extends BaseEventController {
...
+    };
+    const plugins = [];
+    await Promise.all(constants.PLUGINS.map(async (pluginType) => {
+      const plugin = await createPlugin(pluginType);
+      if (plugin) {
+        plugins.push(plugin);
+        parsedConfig[`${pluginType}Id`] = plugin?.id;
+        parsedConfig[pluginType] = plugin?.name;
+      }
+    }));
```
yorkie, line 60: How about returning the plugins via `Promise.all`?

packages/daemon/src/app/controller/pipeline.ts
```diff
@@ -40,8 +40,26 @@ export class PipelineController extends BaseEventController {
     const { name, configUri, config } = this.ctx.request.body;
     const parsedConfig = await parseConfig(configUri || config);
     parsedConfig.name = name;
+    // the plugin name could be git/web url, we need the real name, and create plugin object
+    const createPlugin = async (field: string) => {
+      if (parsedConfig[field]) {
+        const pkg = await this.pluginManager.fetch(parsedConfig[field]);
+        return this.pluginManager.findOrCreateByPkg(pkg);
+      } else {
+        return undefined;
```
yorkie, line 49: No need to return undefined here.

packages/core/src/types/plugins.ts
```diff
@@ -245,5 +245,9 @@ export enum PluginStatus {
   /**
    * the plugin installation pending
    */
-  PENDING
+  PENDING,
+  /**
+   * the plugin installation not started
+   */
+  INIT
```
yorkie, line 252: INITIALIZED?

packages/core/src/types/database.ts
```diff
@@ -34,8 +34,9 @@ export enum PipelineStatus {
  * The pipeline database parameters list.
  */
 export type PipelineDBParams = 'dataCollectParams' | 'dataAccessParams' | 'dataProcessParams' | 'modelDefineParams' | 'modelLoadParams' | 'modelTrainParams' | 'modelEvaluateParams';
+export type PipelineDBId = 'dataCollectId' | 'dataAccessId' | 'dataProcessId' | 'modelDefineId' | 'modelLoadId' | 'modelTrainId' | 'modelEvaluateId';
```
yorkie, line 37: Is this necessary to specify here? The Id should be the only concept inside daemon?

packages/daemon/src/service/plugin.ts
```diff
@@ -123,7 +126,9 @@ export class PluginManager {
         category: pkg.pipcook.category,
         datatype: pkg.pipcook.datatype,
         dest: pkg.pipcook.target.DESTPATH,
-        status: PluginStatus.INSTALLING
+        from: pkg.pipcook.source.from,
+        uri: pkg.pipcook.source.uri,
```
yorkie, line 130: How about adding a source prefix for `uri` and `from`?

packages/daemon/src/app/controller/pipeline.ts
```diff
@@ -45,19 +45,16 @@ export class PipelineController extends BaseEventController {
...
+    const plugins = (await Promise.all(constants.PLUGINS.map(async (pluginType) => {
       const plugin = await createPlugin(pluginType);
       if (plugin) {
-        plugins.push(plugin);
         parsedConfig[`${pluginType}Id`] = plugin?.id;
         parsedConfig[pluginType] = plugin?.name;
+        return plugin;
       }
-    }));
+    }))).filter((plugin) => plugin);
```
yorkie, line 57: It's good to exactly define a function `const isNotUndefined = (f) => f !== undefined` for this?

packages/core/src/types/database.ts
yorkie: Forget to remove it?

packages/core/src/types/database.ts
```diff
@@ -34,8 +34,9 @@ export enum PipelineStatus {
...
 
 /**
  * The pipeline database schema.
  */
-export type PipelineDB = Partial<Record<'id' | 'name' | PluginTypeI | PipelineDBParams, string>>;
+export type PipelineDB = Partial<Record<'id' | 'name' | PluginTypeI | PipelineDBParams | PipelineDBId, string>>;
```
yorkie: Ditto

packages/daemon/src/app/controller/pipeline.ts
yorkie:
```js
const plugins = (await Promise.all(constants.PLUGINS.map(createPlugin)))
  .filter(isNotUndefined)
  .map((plugin) => {
    parsedConfig[`${pluginType}Id`] = plugin?.id;
    parsedConfig[pluginType] = plugin?.name;
    return plugin
  });

Comment (yorkie) [reply packages/daemon/src/app/controller/pipeline.ts:57]: Otherwise LGTM :)

**Comment by yorkie** [comment]
The last CI prints the following error:

```
  pipeline api.job test
    ✓ prepare
    ✗ create pipeline
      - Error: SQLITE_BUSY: database is locked
    ✗ create job
      - TypeError: Cannot read property 'id' of undefined
    ✗ query job info
      - TypeError: Cannot read property 'id' of undefined
    ✗ list job infos
      - TypeError: Cannot read property 'id' of undefined
    ✗ clean
      - TypeError: Cannot read property 'status' of undefined
```

**Comment by FeelyChau** [comment]
> The last CI prints the following error:
> 
> ```
>   pipeline api.job test
>     ✓ prepare
>     ✗ create pipeline
>       - Error: SQLITE_BUSY: database is locked
>     ✗ create job
>       - TypeError: Cannot read property 'id' of undefined
>     ✗ query job info
>       - TypeError: Cannot read property 'id' of undefined
>     ✗ list job infos
>       - TypeError: Cannot read property 'id' of undefined
>     ✗ clean
>       - TypeError: Cannot read property 'status' of undefined
> ```

Look like a processor still take up the db file.

**Comment by yorkie** [comment]
> https://github.com/alibaba/pipcook/blob/daemon/create-pipeline/.github/workflows/build.yml#L63

How can we verify the db file is corresponding to a specific version?

**Comment by FeelyChau** [comment]
> > https://github.com/alibaba/pipcook/blob/daemon/create-pipeline/.github/workflows/build.yml#L63
> 
> How can we verify the db file is corresponding to a specific version?

No need to do this, this db file is a baseline, all the migration scripts will be tested.

**Comment by yorkie** [comment]
Ok, I suggest moving this file to the repository and renamed like: "baseline-v1.db".

**Comment by FeelyChau** [comment]
> > https://github.com/alibaba/pipcook/blob/daemon/create-pipeline/.github/workflows/build.yml#L63
> 
> How can we verify the db file is corresponding to a specific version?

Maybe we should rename the file to `pipcook-base.db` in OSS to make it more specific.

**Comment by yorkie** [comment]
> > > https://github.com/alibaba/pipcook/blob/daemon/create-pipeline/.github/workflows/build.yml#L63
> > 
> > 
> > How can we verify the db file is corresponding to a specific version?
> 
> Maybe we should rename the file to `pipcook-base.db` in OSS to make it more specific.

1. We use OSS to save the model and dataset because they cost too many spaces really, but the *.db seems not too big that, so IMO it should be saved to daemon/test/fixtures for running the tests without internet.
2. We don't promise the backward-compatible of the database for major upgrades, the filename with version might make sense.

**Comment by FeelyChau** [comment]
> > > > https://github.com/alibaba/pipcook/blob/daemon/create-pipeline/.github/workflows/build.yml#L63
> > > 
> > > 
> > > How can we verify the db file is corresponding to a specific version?
> > 
> > 
> > Maybe we should rename the file to `pipcook-base.db` in OSS to make it more specific.
> 
> 1. We use OSS to save the model and dataset because they cost too many spaces really, but the *.db seems not too big that, so IMO it should be saved to daemon/test/fixtures for running the tests without internet.
> 2. We don't promise the backward-compatible of the database for major upgrades, the filename with version might make sense.

Done.

File: .github/workflows/build.yml [hunk 9]
```diff
@@ -60,6 +60,7 @@ jobs:
         npm test
     - name: daemon start
       run: |
+        mkdir -p ~/.pipcook/db && cp ./packages/daemon/test/res/baseline-1.0.0.db ~/.pipcook/db/pipcook.db
```
yorkie, line 63: .
```suggestion
        mkdir -p ~/.pipcook/db && cp ./packages/daemon/test/res/baseline-v1.db ~/.pipcook/db/pipcook.db
```

yorkie: Shall we have a restore progress for later steps?

yorkie (approved):
Looks good to me

.github/workflows/build.yml
```diff
@@ -58,6 +58,9 @@ jobs:
     - name: Running unit tests
       run: |
         npm test
+    - name: migration
+      run: |
+        [ "${{ runner.os }}" -eq "Linux" ] && mkdir -p ~/.pipcook/db && cp ./packages/daemon/test/res/baseline-v1.db ~/.pipcook/db/pipcook.db || true
```
yorkie, line 63: This line is too long :(

yorkie (approved):

WenheLI (approved):
LGTM

rickycao-qy (approved):
LGTM

PR #518, opened 2020-08-27
costa: remove dependecies when uninstall plugins

FeelyChau:


yorkie:
The test is needed :)

FeelyChau:
> The test is needed :)

Done.

packages/costa/test/runtime_test.ts
```diff
@@ -1,55 +1,88 @@
...
+    await costa.install(npmPkg, process);
+    // make sure js packages are installed.
+    await stat(path.join(costa.options.installDir, 'node_modules', npmPkg.name));
     // make sure python caches are used.
     await stat(path.join(costa.options.installDir, '.pip/selfcheck.json'));
   });
 
   it('should fetch a plugin and install from tarball', async () => {
-    const collectCsvWithSpecificVer = await costa.fetch('https://registry.npmjs.org/@pipcook/plugins-csv-data-collect/-/plugins-csv-data-collect-0.5.8.tgz');
+    try {
```
yorkie, line 73: This needs to be reverted.

packages/costa/test/runtime_test.ts
```diff
@@ -69,48 +102,24 @@ describe('create a costa runtime', () => {
     expect(collectCsvOnBare.version).toBe(collectCsvLatest.version);
   });
 
-  it('should install the package without conda packages', async () => {
```
yorkie, line 72: What's the reason to remove this case?

packages/costa/test/runtime_test.ts
```diff
@@ -69,48 +102,24 @@ describe('create a costa runtime', () => {
...
-    await costa.install(collectCsv, process);
-    await stat(path.join(
-      costa.options.installDir,
-      'node_modules',
-      collectCsv.name
-    ));
-  });
-
-  it('should install the package with conda packages', async () => {
```
yorkie, line 81: Ditto, we need a reason to remove this case.

packages/costa/test/runtime_test.ts
FeelyChau: https://github.com/alibaba/pipcook/pull/518/files#diff-bdea6cb8b04e910bac8851b60725bbaeL17
This case has tested `install the package without conda packages`.

packages/costa/test/runtime_test.ts
FeelyChau: https://github.com/alibaba/pipcook/pull/518/files#diff-bdea6cb8b04e910bac8851b60725bbaeL30

packages/costa/test/runtime_test.ts
FeelyChau: Updated.

yorkie (approved):
Looks good to me

PR #521, opened 2020-08-27
daemon: refactor database model

FeelyChau:


packages/daemon/src/app.ts
```diff
@@ -0,0 +1,8 @@
+import DB from './boot/database';
+
+module.exports = app => {
+  app.beforeStart(async () => {
+    const db = (await app.applicationContext.getAsync('pipcookDB')) as DB;
+    db.connect();
```
yorkie, line 6: What's the reason to move the db connection here instead of @init decorator?

FeelyChau: @init decorator depend on the dependencies of IOC object, but now, the model apis are offered by static class function like `Pipeline.findAll`, there are no dependencies between Model and Service, so the `DB.connect` never be called though @init.

FeelyChau: And, before this PR, the Models are initialized and call `model.sync` every request, it's unnecessary.

yorkie: Ok, reasonable, thanks.

WenheLI (approved):
LGTM

yorkie:
@FeelyChau May I ask you to share the difference before this PR?

FeelyChau:
> @FeelyChau May I ask you to share the difference before this PR?

Updated.

yorkie (approved):

packages/daemon/src/boot/database.ts
```diff
@@ -30,6 +32,11 @@ export default class DB {
       logging: this.logger
     });
     await sequelize.sync();
+    await Promise.all([
+      pluginModel(sequelize),
+      jobModel(sequelize),
+      pipelineModel(sequelize)
```
yorkie, line 38: Shall we load and initialize these models dynamically?

yorkie:
I found we still need to define the database schema and the entity class, shall we have a way to unify them?

FeelyChau:
> I found we still need to define the database schema and the entity class, shall we have a way to unify them?

I think it's hard to unify them, but we can put the operation `as` into the model, so they could be maintained together.

packages/daemon/src/boot/database.ts
FeelyChau: For such a basic component, the models doesn't seem necessary to load dynamically.

packages/daemon/test/service/pipeline.test.ts
```diff
@@ -1,49 +1,76 @@
...
-      modelDefine: 'modelDefine',
-      modelTrain: 'modelTrain',
-      modelEvaluate: 'modelEval'
-    });
+    const obj = await pipeline.createPipeline(mockPipeline);
+    const job = await pipeline.createJob(obj.id);
+    const jobx = await pipeline.getJobById(job.id);
+    console.log(jobx);
+  });
+  // it('#create pipeline and get created pipeline', async () => {
```
yorkie, line 46: How about using `it.skip`?

FeelyChau: This should not be commented out, reverted.

WenheLI (approved):

yorkie (approved):

PR #522, opened 2020-08-28
cli: fix the commander args for --output

yorkie:
This is caused by an incorrect set to commander's option for `--output`.


FeelyChau (approved):
LGTM.

PR #524, opened 2020-08-28
daemon: check plugins status before response

FeelyChau:


yorkie:
@FeelyChau May I ask you to add a test for this patch?

FeelyChau:
> @FeelyChau May I ask you to add a test for this patch?

Test cases added.

packages/daemon/test/controller/job.test.ts
```diff
@@ -0,0 +1,120 @@
...
+    });
+    app.mockClassFunction('pipelineService', 'fetchPlugins', async (pipeline: any): Promise<any[]> => {
+      assert.deepEqual(pipeline, mockPipeline);
+      throw createHttpError(HttpStatus.NOT_FOUND, 'mock error');
+    });
+    app.mockClassFunction('pipelineService', 'createJob', async (pipelineId: string): Promise<any> => {
+      assert.equal(pipelineId, 'id');
+      return mockJob;
+    });
+    app.mockClassFunction('pipelineService', 'runJob', async (job: any, pipeline: any, plugins: any[], log: any): Promise<any> => {
```
yorkie, line 105: What's the strategy of restoring the app?

FeelyChau: I can't find the document for this, but as I know, the services and controllers in mock cases are the same as IOC lifecycle.

FeelyChau: We should be careful when mock a singleton object.

yorkie (approved):

PR #525, opened 2020-08-28
doc: add `pipcook.params` in spec documentation

yorkie:
I will update to Chinese version when the English LGTY :)

### DIFF
--- a/docs/spec/plugin.md
+++ b/docs/spec/plugin.md
@@ -98,13 +98,13 @@ To declare the loss function to be used in a model plugin:
 {
   "name": "loss",
   "type": "string[]",
-  "options": {
+  "options": [
     "meanSquaredError",
     "meanAbsoluteError",
     "categoricalCrossentropy",
     "sparseCategoricalCrossentropy",
     "binaryCrossentropy"
-  },
+  ],
   "defaultValue": [ "categoricalCrossentropy" ]
 }
 ```
@@ -116,7 +116,7 @@ To declare the language mode to a NLP-related plugin:
   "name": "mode",
   "type": "string[1]",
   "options": [ "cn", "en" ],
-  "defaultValue": "cn"
+  "defaultValue": "cn",
   "description": "Chinese text classification or English text classification, the value can be en or cn"
 }
 ```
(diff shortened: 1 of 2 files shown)


yorkie:
Ping @WenheLI @FeelyChau

FeelyChau:
How could I define a `string`/`number` type with options? I think it is common.
Does this work?
```json
{
  "name": "log-level",
  "type": "string",
  "options": [
    "debug",
    "info",
    "warn",
    "error"
  ],
  "defaultValue": "error",
  "description": "log level value"
}
```

yorkie:
> How could I define a `string`/`number` type with options? I think it is common.
> Does this work?
> 
> ```json
> {
>   "name": "log-level",
>   "type": "string",
>   "options": [
>     "debug",
>     "info",
>     "warn",
>     "error"
>   ],
>   "defaultValue": "error",
>   "description": "log level value"
> }
> ```

The option doesn't work with literal type, because it could be rendered in text input only, and some parameter with option is expected to be a `<select />`.

yorkie:
The correct way is to use the type `string[1]`.

yorkie:
Any ideas? @FeelyChau @WenheLI @rickycao-qy

WenheLI:
> > How could I define a `string`/`number` type with options? I think it is common.
> > Does this work?
> > ```json
> > {
> >   "name": "log-level",
> >   "type": "string",
> >   "options": [
> >     "debug",
> >     "info",
> >     "warn",
> >     "error"
> >   ],
> >   "defaultValue": "error",
> >   "description": "log level value"
> > }
> > ```
> 
> The option doesn't work with literal type, because it could be rendered in text input only, and some parameter with option is expected to be a `<select />`.



@yorkie I am confused here. Why we should use `string[1]` in the above example, as the `log-level` is just a single `string` (Why use an array?

yorkie:
> @yorkie I am confused here. Why we should use string[1] in the above example, as the log-level is just a single string (Why use an array?

As I mentioned before, the string is a literal type that will not have the options and the string array will have the `options` to be configured to a <select />.

WenheLI (approved):
LGTM

rickycao-qy (approved):
LGTM

FeelyChau (approved):
LGTM.

PR #526, opened 2020-08-28
daemon: add job status change event

FeelyChau:


packages/daemon/src/app/controller/base.ts
```diff
@@ -33,6 +33,9 @@ export class BaseEventController extends BaseController {
       logStream.on('error', err => {
         sse.emit('error', err.message);
       });
+      logStream.on('jobStatusChange', data => {
```
yorkie, line 36: This event name looks weird, the log shouldn't be related to the job.

packages/daemon/src/service/log-manager.ts
```diff
@@ -2,8 +2,15 @@ import { Transform, TransformCallback } from 'stream';
 import { open, close, write } from 'fs-extra';
 import { provide, scope, ScopeEnum } from 'midway';
 import { StringDecoder } from 'string_decoder';
-import { generateId } from '@pipcook/pipcook-core';
+import { generateId, PipelineStatus, PluginTypeI } from '@pipcook/pipcook-core';
 
+export type PipcookEventType = 'log' | 'jobStatusChange';
```
yorkie, line 7: I think in log-manager, we should define something more common instead of an event to be related to a job.

yorkie (changes requested):

packages/daemon/src/service/log-manager.ts
FeelyChau: Updated.

packages/daemon/src/app/controller/base.ts
FeelyChau: Updated.

FeelyChau:
Refactor the log-manager, may I ask you to review again please? @yorkie

packages/daemon/src/service/trace-manager.ts
```diff
@@ -1,19 +1,59 @@
...
+export interface JobStatusChangeEvent {
+  jobStatus: PipelineStatus;
+  step?: PluginTypeI;
+  stepAction?: 'start' | 'end';
+}
 /**
  * Log obejct for plugin installing and pipeline running.
  */
-export interface LogObject {
+export class Tracer {
```
yorkie, line 18: How about extends from EventEmitter?

packages/daemon/src/service/trace-manager.ts
```diff
@@ -1,19 +1,59 @@
...
+  emitter: EventEmitter;
+
+  constructor(fdOut: number, fdErr: number) {
+    this.id = generateId();
+    this.stdout = new LogPassthrough(fdOut);
+    this.stderr = new LogPassthrough(fdErr);
+    this.emitter = new EventEmitter();
+  }
+
+  listenLog(level: 'info' | 'warn', cb: (level: string, message: string) => void): Promise<void> {
```
yorkie, line 35: log might be a event

yorkie:
@FeelyChau The current implementation looks too complicated, may I ask you to have more improvements?

FeelyChau:
> @FeelyChau The current implementation looks too complicated, may I ask you to have more improvements?

Yep, I review the code and feel the same.
Working on it.

packages/daemon/src/service/trace-manager.ts
FeelyChau: I think it's not a good idea for parameter type checking.

FeelyChau:
> @FeelyChau The current implementation looks too complicated, may I ask you to have more improvements?

Done.

FeelyChau:
Bug found: #530

FeelyChau:
> Bug found: #530

Solved. It has nothing to do with this PR. Please review again please @yorkie , thanks!

packages/daemon/src/service/pipeline.ts
```diff
@@ -230,26 +238,34 @@ export class PipelineService {
       // ensure the model dir exists
       await fs.ensureDir(modelPath);
 
+      emitJobEvent(PipelineStatus.RUNNING, 'dataCollect', 'start');
       // run dataCollect to download dataset.
       await runnable.start(plugins.dataCollect.plugin, getParams(plugins.dataCollect.params, {
```
yorkie, line 243: Could you please wrap this function with start/end emits?

packages/daemon/src/service/trace-manager.ts
yorkie: Hmm, it's not related to parameter types, what I just call is to use `super()` instead of `this.emitter`.

packages/daemon/src/service/trace-manager.ts
```diff
@@ -0,0 +1,208 @@
...
+ * tracer for plugin installing and pipeline running.
+ */
+export class Tracer {
+  // log id
+  id: string;
+  // stdout stream for log pipe
+  private stdout: LogPassthrough;
+  // stderr stream for log pipe
+  private stderr: LogPassthrough;
+  // event emitter
```
yorkie, line 32: Please make a consistent style for comments.

packages/daemon/src/service/trace-manager.ts
```diff
@@ -0,0 +1,208 @@
...
+  private emitter: EventEmitter;
+
+  private opts?: LogOptions;
+  private fdOut: number;
+  private fdErr: number;
```
yorkie, line 37: What do the above fields mean?

packages/daemon/src/service/trace-manager.ts
```diff
@@ -0,0 +1,208 @@
...
+  constructor(opts?: LogOptions) {
+    this.id = generateId();
+    this.emitter = new EventEmitter();
+    this.opts = opts;
+  }
+
+  /**
+   * init loggers
+   */
+  async initLogger() {
```
yorkie, line 47: Could you please explain a bit the relationship between logger with trace?

packages/daemon/src/service/trace-manager.ts
```diff
@@ -0,0 +1,208 @@
...
+  }
+  /**
+   * get the loggers
+   */
+  getLogger(): { stdout: LogPassthrough; stderr: LogPassthrough } {
+    return { stdout: this.stdout, stderr: this.stderr };
+  }
+
+  /**
+   * listen event
```
yorkie, line 61: Also, explain what's the event in a trace object.

packages/daemon/src/service/trace-manager.ts
```diff
@@ -0,0 +1,208 @@
...
+    pipeLog('info', this.stdout);
+    pipeLog('warn', this.stdout);
+  }
+
+  /**
+   * emit event to client
+   * @param type event type
+   * @param data only JobStatusChangeEvent for now
+   */
+  emit(type: PipcookEventType, data: JobStatusChangeEvent) {
```
yorkie, line 89: Ok, I got you but here is there necessary to use type here?

yorkie:
Thank you @FeelyChau and please this PR tries to refactor the internal module, so we have to explain more details includes:

- the new class `Trace`.
- the improvement by this refactoring.

yorkie:
And BTW because we will have a stable release next week, so I'm -1 on landing this refactored change, is there possible to just support emiting these events out?

yorkie:
> And BTW because we will have a stable release next week, so I'm -1 on landing this refactored change, is there possible to just support emiting these events out?

Or we need almost full tests before landing this :)

packages/daemon/src/service/trace-manager.ts
FeelyChau: The `type` parameter is used for client to know what the data is.
So, the types of Type and data are bound, and we need to wrap them.

packages/daemon/src/service/trace-manager.ts
FeelyChau: That will expose `emit`/`on` method, IMO, the relationship between the EventEmitter and Tracer is one of inclusion rather than inheritance.
A `Tracer` object **HAS** an event channel, not **IS**, It also has the log stream.

packages/daemon/src/service/trace-manager.ts
FeelyChau: Updated, put the annotation on the class.

packages/daemon/src/service/trace-manager.ts
FeelyChau: Updated, put the annotation on the class.

packages/daemon/src/service/trace-manager.ts
FeelyChau: Done.

packages/daemon/src/service/trace-manager.ts
FeelyChau: Annotation added.

packages/daemon/src/service/trace-manager.ts
FeelyChau: Updated.

packages/daemon/src/service/pipeline.ts
FeelyChau: Done.

FeelyChau:
> And BTW because we will have a stable release next week, so I'm -1 on landing this refactored change, is there possible to just support emiting these events out?

Could we create version branches to avoid this problem?
We could create  branch 1.2-dev for bug fix and 1.3-dev for feature improvement.

packages/daemon/src/app/controller/base.ts
```diff
@@ -2,7 +2,10 @@ import { inject, Context } from 'midway';
 import { ObjectSchema } from 'joi';
 import * as HttpStatus from 'http-status';
 import { ServerSentEmitter } from '../../utils';
-import { LogManager } from '../../service/log-manager';
+import { TraceManager } from '../../service/trace-manager';
+import Debug from 'debug';
+
+const debug = Debug('base.controller');
```
yorkie, line 8: The name is too general, it's recommended to use the package name as a prefix.

packages/daemon/src/service/pipeline.ts
```diff
@@ -218,10 +218,24 @@ export class PipelineService {
...
       }
     };
-
+    const emitJobEvent = (jobStatus: PipelineStatus, step?: PluginTypeI, stepAction?: 'start' | 'end') => {
+      const jobEvent = new JobStatusChangeEvent(
+        jobStatus,
+        step,
+        stepAction
+      );
+      tracer.emit(jobEvent);
```
yorkie, line 227: In the JavaScript world, a emit function is more like to declare as: `emit(name, value)`.

packages/daemon/src/service/pipeline.ts
```diff
@@ -231,25 +245,25 @@ export class PipelineService {
       await fs.ensureDir(modelPath);
 
       // run dataCollect to download dataset.
-      await runnable.start(plugins.dataCollect.plugin, getParams(plugins.dataCollect.params, {
+      await run('dataCollect', plugins.dataCollect.plugin, getParams(plugins.dataCollect.params, {
```
yorkie, line 248: How about `run('dataCollect', ...args)`?

packages/daemon/src/service/trace-manager.ts
```diff
@@ -0,0 +1,250 @@
+import { Transform, TransformCallback } from 'stream';
+import { EventEmitter } from 'events';
+import { open, close, write } from 'fs-extra';
+import { provide, scope, ScopeEnum } from 'midway';
+import { StringDecoder } from 'string_decoder';
+import { generateId, PipelineStatus, PluginTypeI } from '@pipcook/pipcook-core';
+import Debug from 'debug';
+
+const debug = Debug('tracer');
```
yorkie, line 9:
```suggestion
const debug = Debug('daemon.service.tracer');
```

packages/daemon/src/service/trace-manager.ts
```diff
@@ -0,0 +1,250 @@
...
+
+export type PipcookEventType = 'log' | 'job_status';
```
yorkie, line 11:
```suggestion
export type TraceEventType = 'log' | 'job_status';
```

packages/daemon/src/service/trace-manager.ts
```diff
@@ -0,0 +1,250 @@
...
+
+/**
+ * base pipcook event, defined the fields `type` and `data`
+ */
+export class BasePipcookEvent {
```
yorkie, line 16:
```suggestion
export class TraceEvent {
```

yorkie: Maybe interface?

packages/daemon/src/service/trace-manager.ts
```diff
@@ -0,0 +1,250 @@
...
+  data?: any;
+  constructor(public type: PipcookEventType) {}
+}
+
+/**
+ * pipcook event data type for job status change
+ */
+export class JobStatusChangeEvent extends BasePipcookEvent {
```
yorkie, line 24: Could this be defined at JobController?

packages/daemon/src/service/trace-manager.ts
```diff
@@ -0,0 +1,250 @@
...
+    stepAction?: 'start' | 'end';
+    queueLength?: number;
+  };
+  constructor(jobStatus: PipelineStatus, step?: PluginTypeI, stepAction?: 'start' | 'end', queueLength?: number) {
+    super('job_status');
+    this.data = { jobStatus, step, stepAction, queueLength };
+  }
+}
+
+export type LogLevel = 'info' | 'warn' | 'error';
```
yorkie, line 37:
```suggestion
type LogLevel = 'info' | 'warn' | 'error';
```

It's used in this file only.

packages/daemon/src/service/trace-manager.ts
```diff
@@ -0,0 +1,250 @@
...
+  };
+  constructor(level: LogLevel, data: string) {
+    super('log');
+    this.data = { level, data };
+  }
+}
+
+export type PipecookEvent = JobStatusChangeEvent | LogEvent;
+/**
+ * tracer for plugin installing and pipeline running.
```
yorkie, line 54: I think this should be more general.

packages/daemon/src/service/trace-manager.ts
```diff
@@ -0,0 +1,250 @@
...
+ * it has 2 parts: logger and event handler:
+ * logger:
```
yorkie, line 56: Ditto, shall we implement the log parts based on the event?

packages/daemon/src/service/trace-manager.ts
```diff
@@ -0,0 +1,250 @@
...
+ * event handler:
+ * pipe the pipcook event to clients
+ */
+export class Tracer {
+  // trace id
+  id: string;
+  // stdout stream for log pipe
+  private stdout: LogPassthrough;
+  // stderr stream for log pipe
+  private stderr: LogPassthrough;
```
yorkie, line 67: Ditto, in a `Trace` object, maintain `stdout` and `stderr` looks weird a bit, extending or registration would be more reasonable.

packages/daemon/src/service/trace-manager.ts
yorkie: I have commented some, and the logger should be a module(class) based on the Trace.

packages/daemon/src/service/trace-manager.ts
```diff
@@ -0,0 +1,250 @@
...
+  async initLogger() {
+    this.fdOut = this.opts?.stdoutFile ? await open(this.opts?.stdoutFile, 'w+') : -1;
+    this.fdErr = this.opts?.stderrFile ? await open(this.opts?.stderrFile, 'w+') : -1;
+    this.stdout = new LogPassthrough(this.fdOut);
+    this.stderr = new LogPassthrough(this.fdErr);
+  }
+  /**
+   * get the loggers
+   */
+  getLogger(): { stdout: LogPassthrough; stderr: LogPassthrough } {
```
yorkie, line 94: Ditto.

yorkie:
@FeelyChau I think a tracer is more like a connector, which connects to a target object and dispatches some events to clients, for example, someone wants to track the plugin installing logs:

- the related logs might be some trace-able objects, which the tracer will connect to them.
- the tracer object maintains the connections to their clients(from controllers), and dispatch events to them.
- there is a trace-able object called `stdout`, which emits a `data` event, the tracer should promise this event could be sent to clients.

For JobStatusChange, it's similar to the above, we should define a trace-able object for `JobModel`, and dispatch events to clients.

yorkie:
With the above design, the `Trace` could be extended to listen to the other objects in an easy way.

yorkie:
Moreover, we need to add unit tests for `Trace`, too.

packages/daemon/src/service/trace-manager.ts
FeelyChau: Event could't be buffed, if that, we need to implement the event buffer.

packages/daemon/src/service/trace-manager.ts
FeelyChau: I think we should move them out finally, like the `JobResp`/`TraceResp`, because the sdk depends on them too.

packages/daemon/src/service/trace-manager.ts
FeelyChau: Sorry, I doesn't get it. What you mean `the logger should be a module(class) based on the Trace`? In the current implementation, the loggers are part of the tracer.

packages/daemon/src/app/controller/base.ts
FeelyChau: How about `daemon.base-controller`?

packages/daemon/src/service/pipeline.ts
FeelyChau: Or `push`?

packages/daemon/src/service/pipeline.ts
FeelyChau: What's the `args`? If you mean the args of the function `startJob`, it's not correct for function `run`.

packages/daemon/src/service/trace-manager.ts
FeelyChau: Interface will make us to define `type` in every child class.

> Moreover, we need to add unit tests for `Trace`, too.

Tests has been added for tracer and trace manager.

FeelyChau:
> Moreover, we need to add unit tests for `Trace`, too.

Tests has been added for tracer and trace manager.

packages/daemon/src/app/controller/base.ts
```diff
@@ -2,7 +2,10 @@ import { inject, Context } from 'midway';
 import { ObjectSchema } from 'joi';
 import * as HttpStatus from 'http-status';
 import { ServerSentEmitter } from '../../utils';
-import { LogManager } from '../../service/log-manager';
+import { TraceManager } from '../../service/trace-manager';
+import Debug from 'debug';
+
+const debug = Debug('daemon.base-controller');
```
yorkie, line 8: daemon.controller.base

packages/daemon/src/service/pipeline.ts
yorkie: But here should use `emit`, may I ask you to tweak this?

packages/daemon/src/service/pipeline.ts
yorkie:
```js
await run('dataCollect', getParams(plugins.dataCollect.params, { ... });
```

packages/daemon/src/service/trace-manager.ts
```diff
@@ -0,0 +1,251 @@
...
+    level: LogLevel;
+    data: string;
+  };
+  constructor(level: LogLevel, data: string) {
+    super('log');
+    this.data = { level, data };
+  }
+}
+
+export type PipecookEvent = JobStatusChangeEvent | LogEvent;
```
yorkie, line 52: 1. typo: `PipecookEvent` -> `PipcookEvent`
2. I don't think `PipcookEvent` is correct, they're used in daemon only.

packages/daemon/src/service/trace-manager.ts
yorkie: Why not moving this to LogEvent?

packages/daemon/src/service/trace-manager.ts
FeelyChau: SDK will use it too.

yorkie:
@FeelyChau I still found the log-related fields are defined in Tracer class, which should be moved to LogEvent or other log-related classes.

packages/daemon/src/service/trace-manager.ts
FeelyChau: LogEvent is a event entity, how could it handle the stream?

packages/daemon/src/service/pipeline.ts
FeelyChau: But the two parameters are binding, as we discussed above, we should depend on the typings to make it right.

yorkie: Just add an event type for the first argument.

packages/daemon/src/service/trace-manager.ts
```diff
@@ -0,0 +1,251 @@
...
+    level: LogLevel;
+    data: string;
+  };
+  constructor(level: LogLevel, data: string) {
+    super('log');
+    this.data = { level, data };
+  }
+}
+
+export type PipcookEvent = JobStatusChangeEvent | LogEvent;
```
yorkie, line 52: Ditto, I think the PipcookEvent is not proper for this type, it's used in daemon only.

packages/daemon/src/service/trace-manager.ts
yorkie: Or another type or class?

packages/daemon/src/service/trace-manager.ts
yorkie: > the loggers are part of the tracer.

This looks weird, why does the tracer own the log-related props?

packages/daemon/src/service/trace-manager.ts
```diff
@@ -0,0 +1,251 @@
...
+    };
+    pipeLog('info', this.stdout);
+    pipeLog('warn', this.stderr);
+  }
+
+  /**
+   * emit event to client
+   * @param event pipcook event data
+   */
+  emit(event: PipcookEvent) {
```
yorkie, line 127: We may use `dispatchEvent(event)` instead of `emit`, see https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/dispatchEvent.

packages/daemon/src/service/trace-manager.ts
```diff
@@ -0,0 +1,251 @@
...
+@provide('traceManager')
+export class TraceManager {
+  tracerMap = new Map<string, Tracer>();
+
+  /**
+   * create a log object, must call the destory function to clean it up.
+   */
+  async create(opts?: TraceOptions): Promise<Tracer> {
+    const tracer: Tracer = new Tracer(opts);
+    await tracer.initLogger();
```
yorkie, line 223: How about moving the stdout/stderr into this class?

packages/daemon/src/service/trace-manager.ts
```diff
@@ -173,13 +166,26 @@ class LogPassthrough extends Transform {
   decoder = new StringDecoder('utf8');
   last: string;
   fd: number;
-  filename: string;
-  constructor(fd: number) {
+  initialized: boolean;
+  constructor(private filename: string) {
```
yorkie, line 170:
```suggestion
  constructor(filename: string) {
```

yorkie (dismissed):
LGTM if CI is green

yorkie (approved):

rickycao-qy (approved):
LGTM

WenheLI (approved):
LGTM

PR #527, opened 2020-08-28
sdk: add new api to generate the output url to download

yorkie:


packages/sdk/test/job.test.ts
```diff
@@ -0,0 +1,11 @@
+import { PipcookClient } from '../src';
+
+describe('test job apis', () => {
+  const pipcook = new PipcookClient();
+
+  it('should call get pipeline config by id', async () => {
```
FeelyChau, line 6:
```suggestion
  it('should call get output download url by id', async () => {
```

FeelyChau (changes requested):

FeelyChau (approved):

PR #531, opened 2020-08-28
daemon: check log file before reading

FeelyChau:


yorkie:
Throwing an error if log files not exist is reasonable, is there a special use-case that we need to ignore?

packages/cli/src/service/job.ts
```diff
@@ -48,7 +48,7 @@ export async function log(id: string, opts: any): Promise<void> {
   const client = initClient(opts.hostIp, opts.port);
   try {
     const log = await client.job.log(id);
-    logger.info(log);
+    logger.info(JSON.stringify(log));
```
yorkie, line 51: Why change this?

FeelyChau: A bug here, the parameter of `logger.info` should be `string`.
Test case is on the way.

yorkie (approved):

PR #538, opened 2020-08-31
cli: add command "plugin info"

FeelyChau:


yorkie:
@FeelyChau could you please explain why do we add this?

FeelyChau:
User couldn't fetch the plugin detail.

yorkie:
The plugin info is just the plugin's `package.json`, which they could found from the source.

FeelyChau:
> The plugin info is just the plugin's `package.json`, which they could found from the source.

It's incomplete, what we also need: installing time, where the plugin come from, error message if failed, they can't be fetched from package meta.

yorkie:
Reasonable, and the documentation is also needed :)

yorkie (approved):

yorkie (approved):

PR #539, opened 2020-08-31
daemon: fix log missing

FeelyChau:
If no one subscribes the `data` event, the callback will not be called. See:
https://github.com/nodejs/node/blob/master/lib/_stream_transform.js#L37

### DIFF
--- a/packages/daemon/src/service/log-manager.ts
+++ b/packages/daemon/src/service/log-manager.ts
@@ -32,9 +32,6 @@ class LogPassthrough extends Transform {
   }
 
   _transform(chunk: any, encoding: string, callback: TransformCallback): void {
-    if (this.fd > 0) {
-      write(this.fd, chunk);
-    }
     if (this.last === undefined) {
       this.last = '';
     }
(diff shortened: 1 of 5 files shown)


packages/daemon/src/service/log-manager.ts
```diff
@@ -57,6 +54,17 @@ class LogPassthrough extends Transform {
     callback();
   }
 
+  /**
+   * cover Transform.write, otherwise if no `data` event listener,
+   * the callback `_transform` will not be called, but we need to save the log to file.
+   * @param args args for stream.write
```
yorkie, line 60: This description is incorrect, rest parameters should not be supported by typedoc.

packages/daemon/src/service/log-manager.ts
```diff
@@ -57,6 +54,17 @@ class LogPassthrough extends Transform {
...
+   */
+  write(...args: any[]): boolean {
```
yorkie, line 62: Shall we have a unit test for this change?

FeelyChau:
Test added.

FeelyChau:
Crushed when test.
```sh
2020-09-02 13:28:50,015 ERROR 45547 nodejs.unhandledRejectionError: EBADF: bad file descriptor, write
unhandledRejectionError: EBADF: bad file descriptor, write
errno: -9
code: "EBADF"
syscall: "write"
name: "unhandledRejectionError"
pid: 45547
hostname: MacBook-Pro.local
```

yorkie (approved):

rickycao-qy (approved):
LGTM

Issue #520, opened 2020-08-27
build: dev-build doesn't work


FeelyChau:
It's not work for me.
After `npm run dev-build`, I got an error as:
```shell
$ node ./packages/daemon/bootstrap.js
ERROR: Error reading "dist/config/config.js". Error: SyntaxError: Cannot use import statement outside a module
npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! @pipcook/daemon@1.1.2 migration: `sequelize-cli db:migrate`
npm ERR! Exit status 1
npm ERR! 
npm ERR! Failed at the @pipcook/daemon@1.1.2 migration script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.

npm ERR! A complete log of this run can be found in:
npm ERR!     /path/to/.npm/_logs/2020-08-27T16_17_29_953Z-debug.log
(node:16127) UnhandledPromiseRejectionWarning: Error: Command failed: npm run migration
ERROR: Error reading "dist/config/config.js". Error: SyntaxError: Cannot use import statement outside a module
npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! @pipcook/daemon@1.1.2 migration: `sequelize-cli db:migrate`
npm ERR! Exit status 1
npm ERR! 
npm ERR! Failed at the @pipcook/daemon@1.1.2 migration script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.

npm ERR! A complete log of this run can be found in:
npm ERR!     /path/to/.npm/_logs/2020-08-27T16_17_29_953Z-debug.log

    at checkExecSyncError (child_process.js:630:11)
    at execSync (child_process.js:666:15)
    at bootstrap (/path/to/pipcook/packages/daemon/bootstrap.js:48:17)
    at Object.<anonymous> (/path/to/pipcook/packages/daemon/bootstrap.js:97:3)
    at Module._compile (internal/modules/cjs/loader.js:1138:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1158:10)
    at Module.load (internal/modules/cjs/loader.js:986:32)
    at Function.Module._load (internal/modules/cjs/loader.js:879:14)
    at Function.executeUserEntryPoint [as runMain] (internal/modules/run_main.js:71:12)
    at internal/main/run_main_module.js:17:47
[...]

PR #540, opened 2020-09-01
build: fix esBuild only emitting esm format

WenheLI:
This PR fixes the #520 which only generates `esm` format for nodejs.
To make a temporary workaround, I imported `babel` to transform `esm` to `cjs`.
For the build performance, despite using `esbuild` + `babel`, the speed is still out-perform `tsc`:
| tsc | esBuild+Babel | esBuild |
 | - | - | -|
|11.0s | 2.8s | 1.8s |

Note: since `esbuild` omits `interface` in ts, we need to operate a regular `build` prior to `dev-build` to avoid missing `interface` only files.

In the long run, I will add try to add the support of `esm2cjs` transforming for `esBuild` to avoid the babel overhead.


packages/app/package.json
```diff
@@ -10,8 +10,9 @@
     "test": "ts-node ../../run_tests.ts",
     "build": "npm run clean && npm run compile",
     "dev-build": "npm run clean && npm run dev-compile",
+    "dev-trans": "babel --plugins @babel/plugin-proposal-export-namespace-from ./dist/ --out-dir ./dist && babel --plugins @babel/plugin-transform-modules-commonjs ./dist/ --out-dir ./dist",
     "clean": "rm -rf ./dist && rm -rf tsconfig.tsbuildinfo",
-    "dev-compile": "esbuild ./src/*.ts ./src/**/*.ts --tsconfig=tsconfig.json --outdir=dist && sh setup.sh",
+    "dev-compile": "esbuild ./src/*.ts ./src/**/*.ts --tsconfig=tsconfig.json --outdir=dist && npm run dev-trans && sh setup.sh",
```
yorkie, line 15: Shall we have a universal script for wrapping the dev-trans and dev-compile?

WenheLI: For now, this `dev-trans` is just a temporary workaround. We will get rid of it eventually. In this case, I don't think we need to have a universal script.

yorkie:
@WenheLI The current script looks too complicated, may I ask you to simplify this?

WenheLI:
> @WenheLI The current script looks too complicated, may I ask you to simplify this?

esBuild now support directly exporting packages to `cjs`, which means we do not need to manually transform it using `babel`.
See https://github.com/evanw/esbuild/issues/109#event-3734610089, I will update the `esBuild` once it is officially launched.

yorkie:
@WenheLI So the esbuild version update is the only change to fix #520?

WenheLI:
> @WenheLI So the esbuild version update is the only change to fix #520?

Yep, we only need to wait for the official launch.

PR #542, opened 2020-09-01
ci: improve speed

FeelyChau:


.github/workflows/markdown.yml
```diff
@@ -3,15 +3,19 @@ on:
   push:
     branches:
     - master
+    paths:
+    - '**.md'
   pull_request:
     branches:
     - master
+    paths:
+    - '**.md'
```
yorkie, line 12: We can't skip in this way because there are links to other source files.

FeelyChau: How about 'docs/*'?

yorkie (approved):

PR #543, opened 2020-09-01
build: add yarn to manage package install

WenheLI:
This pr adds `yarn` support for our monorepo (along with lerna). 
After a bit investigation, I found it is possible to get rid of lerna if we adapt to yarn2. However, this also introduces other problems we can discuss later.
For now, the yarn can surely improve the installing process:
| npm | yarn |
| - | - |
| 170s | 76s |


package.json
```diff
@@ -23,14 +23,15 @@
...
+    "preinstall": "yarn workspace @pipcook/boa run install-dep",
+    "postinstall": "node ./tools/init_dev.js -f && rm -rf ./node_modules/midway-bin/node_modules/typescript/ && yarn workspace @pipcook/boa run compile",
     "docsify": "npm run typedoc && http-server -c-1 ./docs",
     "lint": "eslint . --ext .ts --ext .js && lerna run --scope @pipcook/daemon lint",
     "pretest": "npm run lint",
     "test": "sh run_tests.sh",
     "test:pipeline": "sh ./run_pipeline.sh",
     "typedoc": "sh tools/mkdoc.sh",
-    "clean": "lerna run clean",
+    "clean": "lerna run clean --parallel --concurrency 10",
```
yorkie, line 34: Shall we use yarn workspace here?

packages/app/package.json
```diff
@@ -12,10 +12,11 @@
     "dev-build": "npm run clean && npm run dev-compile",
     "clean": "rm -rf ./dist && rm -rf tsconfig.tsbuildinfo",
     "dev-compile": "esbuild ./src/*.ts ./src/**/*.ts --tsconfig=tsconfig.json --outdir=dist && sh setup.sh",
-    "compile": "tsc -b tsconfig.json && sh setup.sh"
+    "compile": " tsc -v && tsc -b tsconfig.json && sh setup.sh"
   },
   "license": "Apache 2.0",
   "dependencies": {
+    "crypto": "^1.0.1",
```
yorkie, line 19: Why do we add `crypto` package?

package.json
WenheLI: Yarn workspace does not support parallel mode at this moment, therefore, to speedup those procedures that could run in parallel we should use `lerna`.

yorkie: How about using https://www.npmjs.com/package/concurrently?

package.json
```diff
@@ -16,21 +16,23 @@
...
+    "crypto": "^1.0.1",
     "ts-node": "^8.6.2",
     "typedoc": "^0.17.8",
     "typescript": "^3.9.7"
   },
   "scripts": {
     "build": "lerna run compile",
     "dev-build": "lerna run dev-compile --parallel --concurrency 10",
-    "postinstall": "lerna clean --yes && lerna bootstrap && node ./tools/init_dev.js -f",
+    "preinstall": "yarn workspace @pipcook/boa run install-dep",
```
yorkie, line 27: Could we move this to boa package?

package.json
```diff
@@ -16,21 +16,23 @@
...
+    "postinstall": "node ./tools/init_dev.js -f && rm -rf ./node_modules/midway-bin/node_modules/typescript/ && yarn workspace @pipcook/boa run compile",
```
yorkie, line 28: Ditto.

package.json
```diff
@@ -16,21 +16,23 @@
     "jasmine": "^3.5.0",
     "jasmine-spec-reporter": "^5.0.2",
     "lerna": "^3.18.4",
+    "crypto": "^1.0.1",
```
yorkie, line 19: Which is the dependent of crypto?

appveyor.yml
```diff
@@ -3,7 +3,6 @@ environment:
     - nodejs_version: '8'
     - nodejs_version: '10'
     - nodejs_version: '12'
-    - nodejs_version: '14'
```
yorkie, line 6: This might be unnecessary to be changed in this PR.

.travis.yml
```diff
@@ -1,10 +1,9 @@
-sudo: false
+
 language: node_js
 node_js:
   - '8'
   - '10'
   - '12'
-  - '14'
```
yorkie, line 7: This might be unnecessary to be changed in this PR.

yorkie:
And where do we install yarn?

packages/app/package.json
```diff
@@ -12,7 +12,7 @@
     "dev-build": "npm run clean && npm run dev-compile",
     "clean": "rm -rf ./dist && rm -rf tsconfig.tsbuildinfo",
     "dev-compile": "esbuild ./src/*.ts ./src/**/*.ts --tsconfig=tsconfig.json --outdir=dist && sh setup.sh",
-    "compile": "tsc -b tsconfig.json && sh setup.sh"
+    "compile": " tsc -v && tsc -b tsconfig.json && sh setup.sh"
```
yorkie, line 15: This should be reverted.

packages/daemon/bootstrap.js
```diff
@@ -63,6 +63,9 @@ function createPidfileSync(pathname) {
   if (!await pathExists(midwayPathname)) {
     midwayPathname = path.join(__dirname, '../../midway');
   }
+  if (! await pathExists(midwayPathname)) {
```
yorkie, line 66:
```suggestion
  if (!await pathExists(midwayPathname)) {
```

package.json
WenheLI: `yarn` will not trigger `postinstall` & `preinstall` in the internal packages if no additional packages are downloaded, therefore, we need to put the two scripts at the root `package.json`

package.json
WenheLI: removed

appveyor.yml
WenheLI: Just came into my mind, this file should be removed, since we are using gihub to do CI/CD.

.travis.yml
WenheLI: Ditto

WenheLI:
> And where do we install yarn?

Yarn is built-in for github action's container.

yorkie:
> > And where do we install yarn?
> 
> Yarn is built-in for github action's container.

This is not only for GitHub action, our contributor need to install.

package.json
yorkie: This doesn't work with the separate build :(

package.json
WenheLI: If we plan to totally get rid of `lerna`, we could add this. 
But for now, we still have some functionalities requires `lerna` (publishing & versioning). I think the best practice is to leave them here. And move to `concurrently` as the moment we do not need `lerna` anymore.

WenheLI:
> > > And where do we install yarn?
> > 
> > 
> > Yarn is built-in for github action's container.
> 
> This is not only for GitHub action, our contributor need to install.

We can add a document about `yarn`  installation.

package.json
yorkie: But the `clean` is related to build script, and it looks a bit weird that we use yarn to build and lerna to clean files which are built by yarn.

.github/workflows/release.yml
```diff
@@ -55,8 +55,13 @@ jobs:
           git config --global user.name 'pipcook'
           git config --global user.email 'queyue.crk@alibaba-inc.com'
           git add .
+<<<<<<< HEAD
```
yorkie, line 58: Conflicts?

yorkie (changes requested):

WenheLI:
Too many conflicts need to be resolved. 
Will open another PR.

PR #547, opened 2020-09-02
daemon, cli: remove jobs when remove pipeline

FeelyChau:
```sh
$ pipcook pipeline remove 42lw3pir   
? 1 job which belong to the pipeline will be removed too, continue? Yes
ℹ 1 jobs removed.
✔ pipeline 42lw3pir has removed.
$ pipcook pipeline remove all
? 1 job which belong to the pipeline will be removed too, continue? Yes
ℹ 1 jobs removed.
✔ all pipelines removed.
```


.github/workflows/build.yml
```diff
@@ -107,6 +107,9 @@ jobs:
       run: |
         ./packages/cli/dist/bin/pipcook run https://raw.githubusercontent.com/alibaba/pipcook/master/example/pipelines/text-bayes-classification.json
         ls ./output && rm -rf ./output
+    - name: remove pipelines
+      run: |
+        ./packages/cli/dist/bin/pipcook remove all
```
yorkie, line 112: Should be pipcook pipeline remove all?

packages/cli/src/service/pipeline.ts
```diff
@@ -3,6 +3,7 @@ import * as Url from 'url';
 import { PipelineResp, PluginStatusValue, PipelineConfig } from '@pipcook/sdk';
 import { constants, PluginStatus } from '@pipcook/pipcook-core';
 import { readJson } from 'fs-extra';
+import { prompt } from 'inquirer';
```
yorkie, line 6: I think it's not good behavior to write cli logic in service.

.github/workflows/build.yml
FeelyChau: Fixed.

packages/cli/src/service/pipeline.ts
FeelyChau: Updated.

packages/cli/src/bin/pipcook-pipeline.ts
```diff
@@ -41,9 +42,30 @@ program
...
+          type: 'confirm',
+          name: 'remove',
+          message: `${jobs.length} ${jobs.length > 1 ? 'jobs' : 'job'} which belong to the pipeline will be removed too, continue?`,
+          default: false
+        }
+      ]);
+      confirm = answer.remove;
+    }
+    if (confirm) {
+      remove(id, jobs, opts);
```
yorkie, line 66:
```suggestion
      await remove(id, jobs, opts);
```

packages/daemon/src/app/controller/pipeline.ts
```diff
@@ -77,16 +77,24 @@ export class PipelineController extends BaseEventController {
...
   /**
-   * delete pipeline by id
+   * delete pipeline by id, it will remove the jobs which belong to the pipeline.
    */
   @del('/:id')
   public async removeOne() {
     const { id } = this.ctx.params;
+    const jobs = await this.pipelineService.getJobsByPipelineId(id);
+    for (const job of jobs) {
+      await this.pipelineService.removeJobByModel(job);
```
yorkie, line 96: Shall we remove them in parallel?

FeelyChau: Updated.

yorkie (approved):

yorkie (approved):

PR #558, opened 2020-09-05
sdk: add filename and file type for download api

FeelyChau:


packages/sdk/src/base.ts
```diff
@@ -9,7 +9,7 @@ export class BaseApi {
   }
 
   /**
-   * trace event
+   * trace event, after `Job.run()`, `Pipeline.install()`, `Plugin.createByName()`, `Plugin.createByTarball()` we can get the `traceId` for tracing the progress event.
```
yorkie, line 12: This line is too long :)

packages/sdk/src/interface.ts
```diff
@@ -156,3 +156,10 @@ export interface PipelineConfig {
   name: string;
   plugins: PipelinePlugins;
 }
+
+export type FileDownloadResp = {
+  fileName: string;
```
yorkie, line 161:
```suggestion
  filename: string;
```

packages/sdk/src/interface.ts
```diff
@@ -156,3 +156,10 @@ export interface PipelineConfig {
...
+  fileType: string;
```
yorkie, line 162:
```suggestion
  mimeType: string;
```

packages/sdk/src/job.ts
```diff
@@ -70,8 +70,18 @@ export class Job extends BaseApi {
...
-  downloadOutput(id: string): Promise<FileDownloadResp> {
-    return getFile(`${this.route}/${id}/output`);
+  async downloadOutput(id: string): Promise<FileDownloadResp> {
+    const resp = await getFile(`${this.route}/${id}/output`);
+    // header['content-disposition'] value looks like: 'attachment; filename="pipcook-output-u9fo9dlt.tar.gz"'
+    let fileName = resp.headers['content-disposition']?.match(/\"(.*?)\"/g)[0]?.replace(/\"/g, '');
+    fileName = fileName || 'output.tar.gz';
+    let fileType = resp.headers['content-type'];
+    if (fileType && fileType.startsWith('application/')) {
+      fileType = fileType.substr('application/'.length);
```
yorkie, line 80: Because we have used the mime type, so this is no need to call this.

yorkie (changes requested):

packages/sdk/src/base.ts
FeelyChau: Updated.

packages/sdk/src/interface.ts
FeelyChau: Updated.

packages/sdk/src/interface.ts
FeelyChau: Updated.

packages/sdk/src/job.ts
FeelyChau: Updated.

packages/sdk/src/job.ts
```diff
@@ -70,8 +70,13 @@ export class Job extends BaseApi {
    * you should check the job status before downloading
    * @param id job id
    */
-  downloadOutput(id: string): Promise<FileDownloadResp> {
-    return getFile(`${this.route}/${id}/output`);
+  async downloadOutput(id: string): Promise<FileDownloadResp> {
+    const resp = await getFile(`${this.route}/${id}/output`);
+    // header['content-disposition'] value looks like: 'attachment; filename="pipcook-output-u9fo9dlt.tar.gz"'
+    let filename = resp.headers['content-disposition']?.match(/\"(.*?)\"/g)[0]?.replace(/\"/g, '');
```
yorkie, line 76:
```suggestion
    const filename = resp.headers['content-disposition']?.match(/\"(.*?)\"/g)[0]?.replace(/\"/g, '') || 'output.tgz';
```

yorkie: How about using the library: https://github.com/jshttp/content-disposition?

FeelyChau: It looks great!

packages/sdk/src/job.ts
```diff
@@ -70,8 +70,13 @@ export class Job extends BaseApi {
    * you should check the job status before downloading
    * @param id job id
    */
-  downloadOutput(id: string): Promise<FileDownloadResp> {
-    return getFile(`${this.route}/${id}/output`);
+  async downloadOutput(id: string): Promise<FileDownloadResp> {
+    const resp = await getFile(`${this.route}/${id}/output`);
+    // header['content-disposition'] value looks like: 'attachment; filename="pipcook-output-u9fo9dlt.tar.gz"'
+    let filename = ContentDisposition.parse(resp.headers['content-disposition']).parameters['filename'];
+    filename = filename || 'output.tar.gz';
```
yorkie, line 77: How about merge the above 2 expressions?

FeelyChau: Updated.

packages/sdk/src/job.ts
```diff
@@ -70,8 +70,12 @@ export class Job extends BaseApi {
    * you should check the job status before downloading
    * @param id job id
    */
-  downloadOutput(id: string): Promise<FileDownloadResp> {
-    return getFile(`${this.route}/${id}/output`);
+  async downloadOutput(id: string): Promise<FileDownloadResp> {
+    const resp = await getFile(`${this.route}/${id}/output`);
+    // header['content-disposition'] value looks like: 'attachment; filename="pipcook-output-u9fo9dlt.tar.gz"'
+    let filename = ContentDisposition.parse(resp.headers['content-disposition']).parameters['filename'] || 'output.tar.gz';
```
yorkie, line 76:
```suggestion
    const filename = ContentDisposition.parse(resp.headers['content-disposition']).parameters['filename'] || 'output.tar.gz';
```

packages/sdk/src/job.ts
```diff
@@ -1,7 +1,7 @@
-import { get, getFile, del, post, FileDownloadResp } from './request';
+import { get, getFile, del, post } from './request';
 import { BaseApi } from './base';
-import { JobResp, TraceResp, JobListFilter } from './interface';
-
+import { JobResp, TraceResp, JobListFilter, FileDownloadResp } from './interface';
+import * as ContentDisposition from 'content-disposition';
```
yorkie, line 4: Shall we directly import `parse` from 'content-disposition'?

FeelyChau: Yep, it's better.

packages/sdk/src/job.ts
FeelyChau: Fixed.

yorkie (approved):

Issue #503, opened 2020-08-24
sdk: add a fallback function to write custom `onerror` handler


yorkie:
ng-pipboard added a new page https://pipboard.vercel.app/index.html#/connect, which will be redirected when occurring network errors. Now I implemented this logic manually in every call to SDK, that's great if we can just add an error handler when [creating the client](https://github.com/imgcook/pipboard/blob/master/src/utils/common.js):

```js
const pipcook = new PipcookClient(host, port, {
  onerror: (err) => { ... }
});

// or using EventEmitter
pipcook.on('error', (err) => { ... });
```

FeelyChau:
There is no `SDK` anymore in V2, this issue should be closed now.

PR #560, opened 2020-09-07
sdk: add error callback

FeelyChau:
Implement #503 .

### DIFF
--- a/packages/sdk/src/index.ts
+++ b/packages/sdk/src/index.ts
@@ -2,8 +2,8 @@ import { Job } from './job';
 import { Pipeline } from './pipeline';
 import { Plugin } from './plugin';
 import { get } from './request';
-import { VersionsResp, ConfigResp } from './interface';
-import { emitError } from './base';
+import { VersionsResp, ConfigResp, ApiOption } from './interface';
+import { errorHandle } from './base';
 export { JobStatusValue, PluginStatusValue } from './utils';
 export * from './interface';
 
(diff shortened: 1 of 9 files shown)


packages/sdk/src/job.ts
```diff
@@ -10,15 +10,16 @@ export class Job extends BaseApi {
...
+  constructor(url: string, opts?: InitOption) {
+    super(`${url}/job`, opts);
   }
 
   /**
    * list all jobs.
    * @param filter the filter to list the jobs.
    * @returns The jobs list.
    */
+  @errorHandle()
```
yorkie, line 22: I think there is no need to use decorators, we should handle possible errors for all functions.

packages/sdk/src/index.ts
```diff
@@ -17,7 +18,10 @@ export * from './interface';
  * ```
  */
 export class PipcookClient {
-
+  /**
+   * The error handler, if set, the error will be callback instead of throwing.
+   */
+  onError: (err: Error) => void;
```
yorkie, line 24:
```suggestion
  private onerror: (err: Error) => void;
```

packages/sdk/src/index.ts
```diff
@@ -42,23 +46,26 @@ export class PipcookClient {
...
-  constructor(protocolWithHostname = 'http://127.0.0.1', port = 6927) {
+  constructor(protocolWithHostname = 'http://127.0.0.1', port = 6927, opts?: InitOption) {
+    this.onError = opts?.onError;
     this.endpoint = `${protocolWithHostname}:${port}/api`;
-    this.pipeline = new Pipeline(this.endpoint);
-    this.job = new Job(this.endpoint);
-    this.plugin = new Plugin(this.endpoint);
+    this.pipeline = new Pipeline(this.endpoint, opts);
+    this.job = new Job(this.endpoint, opts);
+    this.plugin = new Plugin(this.endpoint, opts);
```
yorkie, line 54: How about sharing the same `BaseApi` in instances of a client?

yorkie (changes requested):

packages/sdk/src/job.ts
FeelyChau: You mean we should do try-catch in every function?

packages/sdk/src/index.ts
FeelyChau: If the onError function is static, these codes could be simplified.

packages/sdk/src/job.ts
yorkie: Exactly, if someone specifies an error callback, he expects to handle all errors.

packages/sdk/src/index.ts
yorkie: This is unrelated to the function `onerror`, the only difference is the route path, they could be maintained at `Pipeline`, `Job`, and `Plugin`.

FeelyChau:
The priority of this feature has been lowered, so this PR is temporarily closed.

FeelyChau:
> The priority of this feature has been lowered, so this PR is closed for now.

PR #578, opened 2020-09-09
Sync Docker Imaga from Dockerhub to Aliyun when published

rickycao-qy:
When the tag changes, the docker image will be built and publish to docker hub and Aliyun registry.


yorkie (approved):

liangzr (approved):
LGTM :)

PR #581, opened 2020-09-09
Add an example pipeline for NER

rickycao-qy:


yorkie:
Could you please append this to [our pipeline workflow](https://github.com/alibaba/pipcook/blob/master/.github/workflows/pipeline.yml#L14)?

yorkie (approved):

Issue #565, opened 2020-09-08
Configurable parameters per job


WenheLI:
Right now, our pipeline is bound with a fixed set of hyper-parameter and this is not convenient to run multi-job with different sets of hyper-parameter. We should allow passing hyper-parameter on a job-scale.  

To implement it:
- We need to add one column in `Job` model: `param` which is a `JSON string`
- We need to add a new API under `CLI`, something like: `pipcook run <jobid> [param-json]`
- This should also connect with `pipboard` for a GUI operation.

yorkie:
> This should also connect with pipboard for a GUI operation.

I will add a new button "Run with parameters".

FeelyChau:
Is this done by PR #620 ？

WenheLI:
Frontend interaction does not get a chance to implement yet

PR #582, opened 2020-09-09
feat: add tvm wasm export 

WenheLI:
This PR adds a wasm export solution for `tensorflow`.
More models' support will come in the future.

To export the model into wasm, you need to `export WASM=1` in the command line to set the env variable.

The above env variable is just a temporary solution. In the final release, we should be able to config the output format per job. Therefore, this support will be added after #565 is landed.


yorkie:
> To export the model into wasm, you need to export WASM=1 in the command line to set the env variable.

I think adding an option from CLI is better than the environment:

```sh
$ pipcook run foobar.json --type wasm|node
```

yorkie:
And I found some .bc files in our source, it might be unstable because they expect to run on specific v8.

docs/INSTALL.md
```diff
@@ -10,6 +10,8 @@ Before starting the installation, please make sure the following environments ar
 - macOS, Linux
 - Node.js 12
 
+**Note:** To use `wasm` output format, you need to manually install [`emsdk`](https://emscripten.org/docs/introducing_emscripten/index.html) and export the `emcc` and `emsdk` to the environmental variable.
```
yorkie, line 13: Shall we include the emcc/emsdk inside Pipcook?

packages/daemon/templates/wasm/predict.js
```diff
@@ -0,0 +1,41 @@
...
+  const output = model.getOutput(0);
+  inputData.copyFrom(input);
+  model.setInput(modelSpec.inputName, inputData);
+  model.run();
+  await ctx.sync();
+  console.log(output.toArray())
+  return output.toArray();
+}
+
+module.exports = predict;
```
yorkie, line 41: EOF

packages/daemon/package.json
```diff
@@ -61,7 +62,8 @@
     "test": "midway-bin test --ts --full-trace",
     "migration": "sequelize-cli db:migrate",
     "migration:undo": "sequelize-cli db:migrate:undo",
-    "benchmark": "node benchmark/bootstrap.js"
+    "benchmark": "node benchmark/bootstrap.js",
+    "postinstall": "python3 -m pip install https://files.pythonhosted.org/packages/67/ff/011f588d54153c5d8ee3841d137acf752933f78eb1e3d1c5ffc6c1cb5a32/pipcook_tvm-0.7.dev1-cp37-cp37m-macosx_10_15_x86_64.whl"
```
yorkie, line 66: Shall we have a PyPi package for this? Installing from this whl causes the following 2 issues:

1. Python upgrade/downgrade not working.
2. mirror index not working.

packages/daemon/binary/preload.js
```diff
@@ -0,0 +1,41 @@
+/*
```
yorkie, line 1: This is created by us? We don't specify the license in this way :(

packages/daemon/src/service/pipeline.ts
```diff
@@ -383,6 +385,66 @@ export class PipelineService {
...
   }
 
+  private _generateWASMOutput(dist: string, opts: GenerateOptions, fileQueue: Array<Promise<void | string>>): void {
+    const relay = boa.import("tvm.relay");
+    const emcc = boa.import("tvm.contrib.emcc")
+    const keras = boa.import("tensorflow.keras");
+    const {dict, open} = boa.builtins();
+
+    // download tvm runtime from oss
+    const tvmjsPromise = execAsync(`wget http://ai-sample.oss-cn-hangzhou.aliyuncs.com/tvmjs/dist/tvmjs.bundle.js`, {cwd: dist});
```
yorkie, line 395: No need to use string template.

yorkie: We need a cache for this bundle.

packages/daemon/src/service/pipeline.ts
```diff
@@ -383,6 +385,66 @@ export class PipelineService {
...
+    fileQueue.push(tvmjsPromise);
+
+    const model = keras.models.load_model(path.join(opts.modelPath, "model.h5"));
```
yorkie, line 398: How about splitting the framework-specific generator to another file?

yorkie (changes requested):
And the CI should be fixed :)

WenheLI:
Await TVM releasement.

yorkie:
> Await TVM releasement.

@WenheLI May I ask you to share which release is to be waited, and what problem(s) to be resolved by this release?

WenheLI:
We are awaiting https://github.com/apache/incubator-tvm/issues/6486 (v0.7).
They will provide binary distribution after this release, which will help us reduce the redundant work during installation.

yorkie:
See the following notes from CPython Documentation:

> if the Python executable is found in /usr/local/bin/python, it will assume that the libraries are in /usr/local/lib/pythonX.Y. (In fact, this particular path is also the “fallback” location, used when no executable file named python is found along PATH.) The user can override this behavior by setting the environment variable PYTHONHOME, or insert additional directories in front of the standard path by setting PYTHONPATH.

Thus the `PYTHONHOME` is an unnecessary set, because if the python executable is /path/to/boa/.miniconda/bin/python, the libraries are at /path/to/boa/.miniconda/lib/pythonX.Y.

/cc @FeelyChau @WenheLI @rickycao-qy

yorkie:
I will remove it after #602 gets merged because this PR might require it as well :)

packages/daemon/src/service/pipeline.ts
```diff
@@ -317,6 +317,66 @@ export class PipelineService {
     return path.join(CoreConstants.PIPCOOK_RUN, id, 'output.tar.gz');
   }
 
+  // private _generateWASMOutput(dist: string, opts: GenerateOptions, fileQueue: Array<Promise<void | string>>): void {
```
yorkie, line 320: @WenheLI @FeelyChau Boa is working with `worker_threads` at #602, this function could be reimplemented with `worker_threads` to avoid blocking the server.

WenheLI:
Too many conflicts, open another PR for this feature

Issue #576, opened 2020-09-08
pipcook run error: socket.timeout: The read operation timed out when download torch


AnselDai:

yorkie:
@AnselDai You could append a mirror parameter `--tuna` to avoid this error.

AnselDai:
I use `--tuna` parameter but the problem still exist. Can i specify the source when download torch?

yorkie:
Sorry @AnselDai, current Pipcook don't have the ability to specify it in the user interface, but we config the index page at https://github.com/alibaba/pipcook/blob/master/packages/cli/src/config.ts#L9, could you please try to change that at your local machine(/usr/local/lib/node_modules/@pipcook/pipcook-cli/lib/config.js)?

yorkie:
And could you please provide the followings:

- logs
- macOS or Linux
- Node.js version
- the country where your PC is located

Thank you!

AnselDai:
root@instance-pppo0843:~/myfile/pipcook_demo# pipcook run pipeline.json --tuna
✔ plugin @pipcook/plugins-object-detection-pascalvoc-data-collect@1.1.0 has already been installed
✔ plugin @pipcook/plugins-coco-data-access@1.1.1 has already been installed
ℹ PYTHON: Looking in indexes: https://pypi.tuna.tsinghua.edu.cn/simple
ℹ PYTHON: Collecting git+https://gitee.com/pipcook/detectron2.git (from -r /root/.pipcook/plugins/conda_envs/@pipcook/plugins-detectron-fasterrcnn-model-define@1.1.0/requirements.txt (line 6))
ℹ PYTHON:   Cloning https://gitee.com/pipcook/detectron2.git to /tmp/pip-req-build-kki_8eby
⚠ NODE: npm WARN plugins@1.0.0 No description
⚠ NODE: npm WARN plugins@1.0.0 No repository field.
⚠ NODE:
ℹ NODE: + @pipcook/plugins-detectron-fasterrcnn-model-define@1.1.0
ℹ NODE: added 2 packages and audited 150 packages in 5.138s
ℹ NODE:
ℹ NODE: 3 packages are looking for funding
ℹ NODE:   run `npm fund` for details
ℹ NODE:
ℹ NODE: found 0 vulnerabilities
ℹ NODE:
ℹ PYTHON:     Complete output from command python setup.py egg_info:
ℹ PYTHON:     Collecting torch==1.4.0
[...]

PR #583, opened 2020-09-09
costa: import pip cache and pyIndex to pip process

FeelyChau:
Fix: #576 .

### DIFF
--- a/packages/costa/src/runtime.ts
+++ b/packages/costa/src/runtime.ts
@@ -363,7 +363,12 @@ export class CostaRuntime {
         '--default-timeout=1000',
         `--cache-dir=${this.options.installDir}/.pip`
       ]);
-      return spawnAsync(`${envDir}/bin/pip3`, args, {}, stdio);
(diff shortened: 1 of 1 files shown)


packages/costa/src/runtime.ts
```diff
@@ -363,7 +363,12 @@ export class CostaRuntime {
         '--default-timeout=1000',
         `--cache-dir=${this.options.installDir}/.pip`
       ]);
-      return spawnAsync(`${envDir}/bin/pip3`, args, {}, stdio);
+      return spawnAsync(`${envDir}/bin/pip3`, args, {
```
yorkie, line 366: Using [`--install-option`](https://pip.pypa.io/en/stable/reference/pip_install/#install-install-option) might be a standard way.

yorkie:
BTW, I'm still looking for a standard way to address #576 in the Python ecosystem.

WenheLI:
This could fix detectron.

yorkie:
What's the progress of this pull request?

FeelyChau:
#574 has been fixed by #586 .

PR #584, opened 2020-09-09
costa: send signal instead of ipc to destroy client

### LINKED ISSUE #561: training failed，训练cancel 之后训练进程没有退出
image

FeelyChau:
Fix #561 .
Costa sends the destroy-request by IPC, if the client is busy, i.e. training loop in boa stack, the request will not be processed util the `nextTick`, so it will not exit immediately. But we can use signal to make sure the client process can receive it when the OS switches the process from sleeping to running.

### DIFF
--- a/packages/costa/src/runnable.ts
+++ b/packages/costa/src/runnable.ts
@@ -161,7 +161,6 @@ export class PluginRunnable {
     });
     this.state = 'idle';
 
-    console.log('resp', resp);
     // return if the result id is provided.
     const id = resp.params[0];
     if (id === 'error') {
--- a/packages/costa/src/runnable.ts
+++ b/packages/costa/src/runnable.ts
@@ -161,6 +161,7 @@ export class PluginRunnable {
     });
     this.state = 'idle';
 
+    console.log('resp', resp);
     // return if the result id is provided.
     const id = resp.params[0];
     if (id === 'error') {
--- a/packages/costa/src/runnable.ts
+++ b/packages/costa/src/runnable.ts
@@ -19,7 +19,7 @@ export class RunnableResponse implements PluginResponse {
 }
 
 // wait 5000ms for chile process finish.
-const waitForEnd = 5000;
+const waitForDestroied = 5000;
 /**
  * The arguments for calling `bootstrap`.
  */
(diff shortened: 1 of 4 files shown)


yorkie:
@FeelyChau Could you please explain a bit why this PR fixes #561?

packages/costa/src/runnable.ts
```diff
@@ -176,7 +176,7 @@ export class PluginRunnable {
       return;
     }
     this.canceled = true;
-    this.send(PluginOperator.WRITE, { event: 'destroy' });
+    this.handle.kill('SIGHUP');
```
yorkie, line 179: SIGTERM?

yorkie: And shall we have a retry with the signal when detecting the costa process is busy?

FeelyChau: Is there any case we need to wait the client process ends itself? If not , signal will be more quick and directly.

FeelyChau: If we need to wait something done, I think we could handle the signal in client and do the ends, it seems more elegantly.

packages/costa/src/runnable.ts
```diff
@@ -18,6 +18,8 @@ export class RunnableResponse implements PluginResponse {
   }
 }
 
+// wait 1000ms for chile process finish.
+const waitForDestroied = 1000;
```
yorkie, line 22:
```suggestion
const waitForDestroyed = 1000;
```

packages/costa/src/runnable.ts
```diff
@@ -47,7 +49,8 @@ export class PluginRunnable {
   private onread: Function | null;
   private onreadfail: Function | null;
   private ondestroyed: Function | null;
-
+  // timer for wait the process to exit itself
+  private termTimer: NodeJS.Timeout;
```
yorkie, line 53:
```suggestion
  private notRespondTimer: NodeJS.Timeout;
```

packages/costa/src/runnable.ts
```diff
@@ -252,6 +260,9 @@ export class PluginRunnable {
    */
   private async afterDestroy(code: number, signal: NodeJS.Signals): Promise<void> {
     debug(`the runnable(${this.id}) has been destroyed with(code=${code}, signal=${signal}).`);
+    if (this.termTimer) {
+      clearTimeout(this.termTimer);
```
yorkie, line 264: this.notRespondTimer = null;

yorkie (changes requested):

yorkie (approved):

rickycao-qy (approved):
Looks good to me

PR #586, opened 2020-09-10
pip install according to order

rickycao-qy:
pip install -r requirements will change the order of dependencies which may causes failures

### DIFF
--- a/packages/costa/src/runtime.ts
+++ b/packages/costa/src/runtime.ts
@@ -364,7 +364,7 @@ export class CostaRuntime {
         `--cache-dir=${this.options.installDir}/.pip`
       ]);
 
-      for (const req of requirements) {
+      for (const pkg of requirements) {
         const installArgs = args.concat([ req ]);
         await spawnAsync(`${envDir}/bin/pip3`, installArgs, {}, stdio);
       }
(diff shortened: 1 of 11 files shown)


packages/costa/src/runtime.ts
```diff
@@ -355,15 +355,19 @@ export class CostaRuntime {
...
       if (opts.pyIndex) {
         args = args.concat([ '-i', opts.pyIndex ]);
       }
       args = args.concat([
         '--default-timeout=1000',
         `--cache-dir=${this.options.installDir}/.pip`
       ]);
-      return spawnAsync(`${envDir}/bin/pip3`, args, {}, stdio);
+
+      for (const req of requirements) {
```
yorkie, line 367:
```suggestion
      for (const pkg of requirements) {
```

yorkie: req sounds like a request :)

rickycao-qy: Sounds good

packages/costa/src/runtime.ts
```diff
@@ -355,15 +355,19 @@ export class CostaRuntime {
...
         args = args.concat([ '-i', opts.pyIndex ]);
       }
       args = args.concat([
         '--default-timeout=1000',
         `--cache-dir=${this.options.installDir}/.pip`
       ]);
-      return spawnAsync(`${envDir}/bin/pip3`, args, {}, stdio);
+
+      for (const pkg of requirements) {
+        const installArgs = args.concat([ req ]);
```
yorkie, line 368:
```suggestion
        const installArgs = args.concat([ pkg ]);
```

yorkie (approved):

PR #588, opened 2020-09-11
Fix/daemon service url

gindis:
fix daemon startup prompt and update document.

### DIFF
--- a/packages/cli/src/bin/pipcook-daemon.ts
+++ b/packages/cli/src/bin/pipcook-daemon.ts
@@ -36,7 +36,7 @@ async function start(): Promise<void> {
     if (message.event === 'ready') {
       daemon.disconnect();
       daemon.unref();
-      logger.success('Pipcook service started, please visit https://pipboard.vercel.app');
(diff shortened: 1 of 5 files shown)


packages/cli/src/bin/pipcook-daemon.ts
```diff
@@ -36,7 +36,7 @@ async function start(): Promise<void> {
     if (message.event === 'ready') {
       daemon.disconnect();
       daemon.unref();
-      logger.success(`Pipcook is on http://localhost:${message.data.listen}`);
+      logger.success('Pipcook service started, please visit https://pipboard.vercel.app');
```
rickycao-qy, line 39: Shall we remain the message of port in case user need to expose the pipcook service? Probably 'Pipcook service started on ${message.data.listen}. To open pipboard, please visit https://pipboard.vercel.app' ?

gindis: sure :)

rickycao-qy (approved):
LGTM

PR #590, opened 2020-09-15
build: use yarn

WenheLI:
For more detail, refer to #543


package.json
```diff
@@ -23,14 +23,14 @@
   "scripts": {
     "build": "lerna run compile",
     "dev-build": "lerna run dev-compile --parallel --concurrency 10",
-    "postinstall": "lerna clean --yes && lerna bootstrap && node ./tools/init_dev.js -f",
+    "postinstall": "node ./tools/init_dev.js -f && rm -rf ./node_modules/midway-bin/node_modules/typescript/",
```
yorkie, line 26: I would prefer to move this to another script and flag it as a bug from midway?

tools/init_dev.js
```diff
@@ -64,7 +65,17 @@ if (process.argv.length > 2 && process.argv[2] === '-f') {
   force = true;
 }
 
-const futures = [];
+// need to manually delete ts under midway since midway does not handle yarn installation properly
+const removeMidwayTS = () => new Promise((resolve, reject) => {
+  const tsPath = path.join(__dirname, "../", "./node_modules/midway-bin/node_modules/typescript/");
+  exec(`rm -rf ${tsPath}`, (err, stdout, stderr) => {
```
yorkie, line 71: Use `util.promisify` instead.

tools/init_dev.js
```diff
@@ -64,7 +65,17 @@ if (process.argv.length > 2 && process.argv[2] === '-f') {
   force = true;
 }
 
-const futures = [];
+// need to manually delete ts under midway since midway does not handle yarn installation properly
+const removeMidwayTS = () => new Promise((resolve, reject) => {
+  const tsPath = path.join(__dirname, "../", "./node_modules/midway-bin/node_modules/typescript/");
```
yorkie, line 70: Shall we have a discussion about this with midway team?

yorkie (changes requested):

tools/init_dev.js
FeelyChau: And we should use single quotes with string.

FeelyChau (changes requested):

tools/init_dev.js
```diff
@@ -64,7 +65,17 @@ if (process.argv.length > 2 && process.argv[2] === '-f') {
   force = true;
 }
 
-const futures = [];
+// need to manually delete ts under midway since midway does not handle yarn installation properly
```
yorkie, line 68: @WenheLI May I ask you to describe which is not handled properly for yarn, thus we can track which could be improved :)

WenheLI:
Miday requires a fixed version (3.6.x) of `tsc` which does not have `optional chain` support. This does not have any impact if we are using `npm`. However, with `yarn`, packages will be flattened and the flattening operation makes the older version of `tsc` available to `midway-bin` which is not the expected behavior. 
One possible solution might be to give an environment variable to define the path of `tsc`.

yorkie (approved):

rickycao-qy (approved):
LGTM

packages/daemon/bootstrap.js
```diff
@@ -63,6 +63,10 @@ function createPidfileSync(pathname) {
   if (!await pathExists(midwayPathname)) {
     midwayPathname = path.join(__dirname, '../../midway');
   }
+  if (!await pathExists(midwayPathname)) {
```
FeelyChau, line 66: What's the condition for?

FeelyChau (changes requested):

FeelyChau (approved):
LGTM.

PR #591, opened 2020-09-15
pipeline: job param

WenheLI:
This PR adds job-level param.
#565

### DIFF
--- a/packages/cli/src/bin/pipcook-job.ts
+++ b/packages/cli/src/bin/pipcook-job.ts
@@ -18,6 +18,7 @@ program
   .option('--output <dir>', 'the output directory name', 'output')
   .option('-h|--host-ip <ip>', 'the host ip of daemon')
   .option('-p|--port <port>', 'the port of daemon')
+  .option('-c|config <config file>', 'the config file for this job')
   .action(runAndDownloadById)
   .description('run a job from a pipeline id');
 
--- a/packages/daemon/bootstrap.js
+++ b/packages/daemon/bootstrap.js
@@ -64,6 +64,9 @@ function createPidfileSync(pathname) {
     midwayPathname = path.join(__dirname, '../../midway');
   }
   if (!await pathExists(midwayPathname)) {
+    midwayPathname = path.join(__dirname, '../../node_modules/midway');
+  }
+  if (!await pathExists(midwayPathname)) {
     throw new TypeError('daemon is not installed correctly.');
   }
   const opts = {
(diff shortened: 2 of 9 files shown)


WenheLI:
Await unit test

yorkie:
> Await unit test

Is the unit test ready now?

WenheLI:
> > Await unit test
> 
> Is the unit test ready now?

The Unit Test is not added yet. I will try to put them together today.

packages/cli/src/bin/pipcook-job.ts
```diff
@@ -18,6 +18,7 @@ program
   .option('--output <dir>', 'the output directory name', 'output')
   .option('-h|--host-ip <ip>', 'the host ip of daemon')
   .option('-p|--port <port>', 'the port of daemon')
+  .option('-c|config <config file>', 'the config file for this job')
```
yorkie, line 21: `pipcook run <id> -c foobar.json` looks like weird, we should only append the parameters only, not the complete config file.

packages/daemon/bootstrap.js
```diff
@@ -63,6 +63,9 @@ function createPidfileSync(pathname) {
   if (!await pathExists(midwayPathname)) {
     midwayPathname = path.join(__dirname, '../../midway');
   }
+  if (!await pathExists(midwayPathname)) {
+    midwayPathname = path.join(__dirname, '../../node_modules/midway');
```
yorkie, line 67: This might be duplicated with #590

yorkie (changes requested):
Before landing this feature, we have to change the behavior to run a job might be:

1. copy the parameters from pipeline.
2. append the extra parameters to the copid parameters.
3. save the parameters to job instance.
4. start running.

yorkie:
To avoid the duplicated definition of parameters, how about just define a string(JSON) for the parameters?

packages/cli/src/bin/pipcook-job.ts
WenheLI: The config here refers to a param config file.

yorkie: I'm confused that, shall we have just command line parameter for that?

WenheLI: The job param might be extremely long. In this case, putting params in CLI may not be the desired way.

yorkie: Hmm, but maintaining 2 configs is too confused for our end-users.

PR #593, opened 2020-09-22
[WIP] CI/CD: add benchmark support 

WenheLI:


yorkie:
How about moving the benchmark CI to the benchmark repo? It could reduce the jobs of the main repo.

WenheLI:
> How about moving the benchmark CI to the benchmark repo? It could reduce the jobs of the main repo.

How should we trigger the Benchmark hook if it is in a separate repo?

yorkie:
> > How about moving the benchmark CI to the benchmark repo? It could reduce the jobs of the main repo.
> 
> How should we trigger the Benchmark hook if it is in a separate repo?

Just noticed that it's triggered for every pull request, how about change it to crontab? Do we need to run benchmarks for every pull request?

WenheLI:
> > > How about moving the benchmark CI to the benchmark repo? It could reduce the jobs of the main repo.
> > 
> > 
> > How should we trigger the Benchmark hook if it is in a separate repo?
> 
> Just noticed that it's triggered for every pull request, how about change it to crontab? Do we need to run benchmarks for every pull request?

This is for testing. My initial design is to run the benchmark on every releasement

yorkie:
Shall we rewrite benchmark scripts in bash at first? Thus we could run benchmark results out in anytime without GitHub Actions.

- able to run benchmarks and upload results without GitHub Actions.
- config the benchmark trigger, which might be discussed later.

WenheLI:
> Shall we rewrite benchmark scripts in bash at first? Thus we could run benchmark results out in anytime without GitHub Actions.
> 
> * able to run benchmarks and upload results without GitHub Actions.
> * config the benchmark trigger, which might be discussed later.

Sure, that sounds promising.

yorkie:
Is this ready for review @WenheLI?

PR #594, opened 2020-09-22
docker: fix cuda environment

FeelyChau:
Fix:
```sh
[tensorflow/stream_executor/platform/default/dso_loader.cc:55] Could not load dynamic library 'libcuda.so.1'; dlerror: libcuda.so.1: cannot open shared object file: No such file or directory; LD_LIBRARY_PATH: /usr/local/cuda-10.1/lib64:/usr/local/cuda-10.2/lib64:/opt/rh/rh-nodejs12/root/usr/lib64
```

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,19 +1,15 @@
-FROM node:latest
+FROM nvidia/cuda:10.1-cudnn7-devel
 ARG VER=latest
 
 LABEL version=${VER}
 LABEL description="docker image for pipcook runtime"
 
-WORKDIR /root/
-RUN npm install @pipcook/pipcook-cli@${VER} -g
-RUN pipcook init ${VER} --tuna
-
-COPY --from=nvidia/cuda:10.2-cudnn7-runtime /usr/local/cuda-10.2 /usr/local/cuda-10.2
-COPY --from=nvidia/cuda:10.1-cudnn7-runtime /usr/local/cuda-10.1 /usr/local/cuda-10.1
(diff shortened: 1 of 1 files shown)


yorkie (approved):

FeelyChau:
I checked the image and in the GPU-enabled container,  there are some issues:
1. cudnn not found
2. detectron2 not work with GPU (maybe the nvcc not found in the path)
We still need to verify it again.

FeelyChau:
Verified on:

* docker 17.06.2-ce
* Nvidia T4.

There is still a issue left:
Before we run a detectron pipeline, we should startup the python and run:` import matplotlib.pyplot `. This will generate the font list file at /root/.cache/matplotlib/fontlist-v330.json, otherwise the pipeline will fail with coredump. Strangely, this command only needs to be executed once, even if the font list file is deleted, the pipeline works well.

It looks like a bug from `matplotlib.pyplot `.

ppwwyyxx:
Yeah the behavior sounds like the same as https://github.com/skvark/opencv-python/issues/381. However inside a docker it's recommended to use `opencv-python-headless` instead so the issue would not exist at the beginning.

yorkie:
Totally agreed with @ppwwyyxx, we should use `opencv-python-headless`.

FeelyChau:
As https://github.com/skvark/opencv-python/issues/381 said, the latest version of opencv should have resolved this issue, we should confirm it first.

FeelyChau:
The plugin dependency has been fix, see
https://github.com/imgcook/pipcook-plugin-detectron-model-train/pull/2
https://github.com/imgcook/pipcook-plugin-detectron-evaluator/pull/2
https://github.com/imgcook/pipcook-plugin-detectron-maskrcnn-model/pull/1
I think we could merge this now.
@yorkie

yorkie (approved):

PR #595, opened 2020-09-22
costa: support multi version of plugins

FeelyChau:
1. support multi version of plugins.
2. use `pnpm` instead of `npm`.

todo list:

* unit tests need to be added.
* script init-dev should be updated.
* boa should be installed into pnpm cache.
* benchmark.


yorkie:
What about the cnpm?

FeelyChau:
> What about the cnpm?

[The `alias` installation is not done for npmInstall](https://github.com/cnpm/npminstall/pull/327), we could make benchmark if it's done .

yorkie:
Because we are always using NPM for plugin installation, so we don't consider the CNPM.

FeelyChau:
Supporting multi versions for plugin need to refactor the database and costa runtime,  it looks like a lot of work, and we can do it in version 2.0.

PR #596, opened 2020-09-23
costa: fix stream block

FeelyChau:
Tarball-upload will block the stream processing if the disk performance is weak or the uploaded file is large.


packages/costa/src/utils.ts
```diff
@@ -30,16 +33,4 @@ export function pipeLog(readable: NodeJS.ReadableStream, writable: NodeJS.Writab
...
- */
-export function pipeGracefully(readStream: NodeJS.ReadableStream, writeStream: NodeJS.WritableStream): Promise<void> {
-  return new Promise((resolve, reject) => {
-    readStream.pipe(writeStream);
-    readStream.on('error', reject);
-    writeStream.on('error', reject);
-    writeStream.on('finish', resolve);
-  });
-}
+export const pipeGracefully = util.promisify(pipeline);
```
yorkie, line 36: We don't need wrapping this anymore.

packages/costa/src/runtime.ts
```diff
@@ -107,13 +107,22 @@ function fetchPackageJsonFromTarball(filename: string): Promise<any> {
...
+  const unzip = createUnzip();
+  const extract = tar.extract();
+  extract.on('entry', (header, stream, next) => {
+    if (header.name === 'package/package.json') {
+      stream.on('data', (buf) => packageJson += buf);
+    }
+    stream.once('end', next);
+    stream.resume();
+  });
+  await pipeGracefully(fileStream, unzip, extract);
```
yorkie, line 124:
```
pipeline(fileStream, unzip, stract);
```

FeelyChau: Updated.

yorkie (approved):

PR #600, opened 2020-09-29
daemon: support to bind host and port

FeelyChau:


packages/cli/src/bin/pipcook-daemon.ts
```diff
@@ -36,7 +37,8 @@ async function start(): Promise<void> {
     if (message.event === 'ready') {
       daemon.disconnect();
       daemon.unref();
-      logger.success(`Pipcook service started on ${message.data.listen}. To open pipboard, please visit https://pipboard.vercel.app`);
+      let serverInfo = (message.data as any).listen ?? `${message.data.host}:${message.data.port}`;
```
yorkie, line 40:
```suggestion
      const serverInfo = (message.data as any).listen ?? `${message.data.host}:${message.data.port}`;
```

packages/daemon/bootstrap.js
```diff
@@ -11,7 +11,8 @@ const PIPCOOK_HOME = os.homedir() + '/.pipcook';
 const DAEMON_PIDFILE = PIPCOOK_HOME + '/daemon.pid';
 const DAEMON_CONFIG = PIPCOOK_HOME + '/daemon.config.json';
 const PIPCOOK_DB = PIPCOOK_HOME + '/db/pipcook.db';
-const PORT = 6927;
+let PORT = 6927;
+let HOST = '127.0.0.1';
```
yorkie, line 15: The default host should be null.

packages/cli/src/bin/pipcook-daemon.ts
yorkie: This looks too complicated.

.github/workflows/build.yml
```diff
@@ -131,3 +131,9 @@ jobs:
     - name: "pipcook plugin-dev"
       run: |
         ./packages/cli/dist/bin/pipcook plugin-dev -t dataCollect
+    - name: "daemon start on customized port"
+      run: |
+        ./packages/cli/dist/bin/pipcook daemon stop
+        echo "{ \"host\": \"127.0.0.1\", \"port\": 7000 }" > ~/.pipcook/daemon.config.json
+        ./packages/cli/dist/bin/pipcook daemon start
+        ./packages/cli/dist/bin/pipcook pipeline list -h 127.0.0.1 -p 7000
```
yorkie, line 139: Should stop daemon, too.

packages/cli/src/bin/pipcook-daemon.ts
```diff
@@ -16,7 +16,8 @@ const BOOTSTRAP_HOME = path.join(DAEMON_HOME, 'bootstrap.js');
 interface DaemonBootstrapMessage {
   event: string;
   data: {
-    listen: number;
```
yorkie, line 19: We should keep it there if we are still using, or create a new interface `DaemonBootstrapV2`

packages/cli/src/bin/pipcook-daemon.ts
```diff
@@ -36,7 +37,8 @@ async function start(): Promise<void> {
     if (message.event === 'ready') {
       daemon.disconnect();
       daemon.unref();
-      logger.success(`Pipcook service started on ${message.data.listen}. To open pipboard, please visit https://pipboard.vercel.app`);
+      const serverInfo = (message.data as any).listen ?? `${message.data.host}:${message.data.port}`;
```
yorkie, line 40: We should split this expression into an `if-else` clause, and add some comments to describe the reason to check.

packages/daemon/bootstrap.js
```diff
@@ -11,7 +11,8 @@ const PIPCOOK_HOME = os.homedir() + '/.pipcook';
 const DAEMON_PIDFILE = PIPCOOK_HOME + '/daemon.pid';
 const DAEMON_CONFIG = PIPCOOK_HOME + '/daemon.config.json';
 const PIPCOOK_DB = PIPCOOK_HOME + '/db/pipcook.db';
-const PORT = 6927;
+let PORT = 6927;
+let HOST = null;
```
yorkie, line 15: I'd like to define variables in the latest function scope.

FeelyChau:
Updated, could you review again please? @yorkie

packages/cli/src/bin/pipcook-daemon.ts
```diff
@@ -36,13 +41,24 @@ async function start(): Promise<void> {
     if (message.event === 'ready') {
       daemon.disconnect();
       daemon.unref();
-      logger.success(`Pipcook service started on ${message.data.listen}. To open pipboard, please visit https://pipboard.vercel.app`);
+      let serverInfo;
+      // TODO(Feely): the listen field is deprecated, we will remove this
```
yorkie, line 45: When do we remove it?

packages/cli/src/bin/pipcook-daemon.ts
```diff
@@ -36,13 +41,24 @@ async function start(): Promise<void> {
...
+      if (message.data.listen) {
+        serverInfo = message.data.listen;
```
yorkie, line 47:
```suggestion
        serverInfo = `http://localhost:${message.data.listen}`;
```

packages/cli/src/bin/pipcook-daemon.ts
FeelyChau: Maybe v2.0.0?

packages/cli/src/bin/pipcook-daemon.ts
```diff
@@ -36,13 +41,24 @@ async function start(): Promise<void> {
...
+      } else if (message.data.host && message.data.port) {
+        serverInfo = `${message.data.host}:${message.data.port}`;
```
yorkie, line 49:
```suggestion
        serverInfo = `http://${message.data.host}:${message.data.port}`;
```

packages/cli/src/bin/pipcook-daemon.ts
```diff
@@ -36,13 +41,24 @@ async function start(): Promise<void> {
...
+      }
+      logger.success(`Pipcook service started on ${serverInfo}. To open pipboard, please visit https://pipboard.vercel.app`);
     }
   });
   daemon.on('exit', async (code: number) => {
     logger.fail(`Pipcook daemon starts failed with code(${code}).`, false);
     // TODO(yorkie): check if this is local mode.
```
yorkie, line 56: Could this be removed now?

packages/daemon/bootstrap.js
```diff
@@ -81,17 +91,21 @@ function createPidfileSync(pathname) {
     });
     // emit `server` event in app
     app.emit('server', server);
-
+  
     // server listen
     await new Promise(resolve => {
-      server.listen(PORT, resolve);
+      if (HOST) {
```
yorkie, line 97:
```suggestion
      if (typeof HOST === 'string') {
```

packages/daemon/bootstrap.js
```diff
@@ -81,17 +91,21 @@ function createPidfileSync(pathname) {
...
+        server.listen(PORT, resolve);
+      }
     });
   } catch (err) {
     return exitProcessWithError(err);
   }
 
   process.title = 'pipcook.daemon';
-  console.info('Server is listening at http://localhost:%s, cost %ss', PORT, process.uptime());
+  console.info('Server is listening at http://%s:%s, cost %ss', HOST || '0.0.0.0', PORT, process.uptime());
```
yorkie, line 108: 0.0.0.0 won't be accessible, it should be 127.0.0.1 :(

yorkie:
Updated @FeelyChau

packages/cli/src/bin/pipcook-daemon.ts
FeelyChau: Because we use cli like:
```sh
$ picpook job list -h localhost -p 6927
```
so, i think `localhost:6927` is better than `http://locahost:6927`.

packages/daemon/bootstrap.js
FeelyChau: `0.0.0.0` is the default value if the host not set, see https://nodejs.org/api/net.html#net_server_listen_port_host_backlog_callback

packages/cli/src/bin/pipcook-daemon.ts
yorkie: This might not be a breaking change, right?

packages/cli/src/bin/pipcook-daemon.ts
yorkie: User can directly just click the `http://locahost:6927` to open it in the browser.

packages/daemon/bootstrap.js
FeelyChau: Actually if the host is number 127.1, it works well on macOS and linux :)
If it must be a string, I think it's better to check when we read it from config file.

packages/cli/src/bin/pipcook-daemon.ts
FeelyChau: We have not implemented the local mode, should we remove it?

yorkie: Keep it there, just resolve it.

packages/daemon/bootstrap.js
yorkie: But 0.0.0.0 is not working with browsers like chrome.

FeelyChau: But 127.0.0.1 is not correct, we bind all the network actually.

yorkie: 127.0.0.1 < 0.0.0.0, and I think we don't need to make sure this address is correct exactly, otherwise, we should list all available local network address instead of 0.0.0.0, which is invalid to access.

FeelyChau: If this, how could we distinguish between '0.0.0.0' and '127.0.0.1', it's confusing.

FeelyChau:
Updated, removed the protocol `http://` from server address.

yorkie (approved):

packages/cli/src/bin/pipcook-daemon.ts
FeelyChau: Yes, if remove this field, the old version of pipcook-cli will not work as expect.

yorkie (approved):

PR #601, opened 2020-09-29
ci: delete daily run of NER pipeline

rickycao-qy:
Currently, the CI machine to run pipeline only has 4G memory. This is not enough for BERT model. This PR is to stop this pipeline's daily run temporarily. For the future, we need to transfer our pipeline run to our own machine or to other CI services


FeelyChau (approved):
We will rollback it when GPU server is ready.

yorkie (approved):

Issue #579, opened 2020-09-09
Daemon should handle npm packages from different sources


rickycao-qy:
We are supporting four different sources of npm packages now: npm registry / git address / url download / local upload.

The problem is that we mix up these different sources and cannot distinguish between them after installed. Pipcook now will only record the package name. So if you want to fetch information about this package, you will by default fetch it from npm registry and it probably does not exist in npm at all.

We need to change the way that when the source is git/url/upload, we save the package, fetch the package info, compare md5 with former packages and install them accordingly. Meanwhile, we need to record the original source.

![image](https://user-images.githubusercontent.com/31037093/93282172-594b0a80-f800-11ea-8387-21b6d4fb453f.png)

FeelyChau:
Daemon has been removed in v2, this issue should be closed now.

PR #605, opened 2020-10-09
daemon: Daemon should handle npm packages from different sources

rickycao-qy:
Refer to #579 

Todo: unit tests


packages/costa/package.json
```diff
@@ -23,6 +23,7 @@
     "better-opn": "^1.0.0",
     "debug": "^4.1.1",
     "fs-extra": "^8.1.0",
+    "md5": "^2.3.0",
```
yorkie, line 26: The `md5` depends on the [crypt](https://github.com/pvorb/node-crypt/blob/master/crypt.js) which is implemented in JavaScript for browser, I'd prefer do use the Node.js crypto module, see https://nodejs.org/dist/latest-v12.x/docs/api/crypto.html#crypto_crypto_createhash_algorithm_options, it supports the algorithm `md5`, too. And it should be more stable and performant.

packages/costa/src/runtime.ts
```diff
@@ -67,13 +68,13 @@ interface InstallOptions {
...
 
-function spawnAsync(command: string, args: string[], opts: SpawnOptions, stdio: LogStdio): Promise<void> {
+function spawnAsync(command: string, args: string[], opts: SpawnOptions, stdio?: LogStdio): Promise<void> {
   return new Promise((resolve, reject) => {
     opts.stdio = [ null, 'pipe', 'pipe' ];
     opts.detached = false;
     const child = spawn(command, args, opts);
-    pipeLog(child.stdout, stdio.stdout, stdio.prefix);
-    pipeLog(child.stderr, stdio.stderr, stdio.prefix);
+    stdio?.stdout && pipeLog(child.stdout, stdio.stdout, stdio.prefix);
```
yorkie, line 76: I think we could just verify it in the function `pipeLog()`.

yorkie (changes requested):
Current there are too much flatten functions, we should refactor it to represent the flow in https://github.com/alibaba/pipcook/issues/579 in better way.

packages/costa/src/runtime.ts
```diff
@@ -128,6 +135,22 @@ async function fetchPackageJsonFromTarballStream(fileStream: NodeJS.ReadableStre
   return JSON.parse(packageJson);
 }
 
+async function fetchMd5(filePath: string): Promise<string> {
+  return new Promise((resolve, reject) => {
```
yorkie, line 139: Is there necessary to use the stream API? It seems that the file might be a small package.json file thus digest() looks simple?

packages/costa/package.json
```diff
@@ -31,6 +31,7 @@
     "@types/debug": "^4.1.5",
     "@types/fs-extra": "^8.0.1",
     "@types/jasmine": "^3.5.7",
+    "@types/md5": "^2.2.0",
```
yorkie, line 34: This should be removed, too :)

packages/costa/src/runtime.ts
```diff
@@ -552,6 +584,11 @@ export class CostaRuntime {
         // TODO(feely): not implement the version
         installDir, `node_modules/${pkg.name}@${pkg.version}`)
     };
+    if (source && source.from !== 'npm') {
```
yorkie, line 587: Source check and assignments should be moved forward the fetch* methods.

packages/costa/src/utils.ts
```diff
@@ -12,6 +12,9 @@ export interface LogStdio {
  * @param prefix the log prefix
  */
 export function pipeLog(readable: NodeJS.ReadableStream, writable: NodeJS.WritableStream, prefix?: string): void {
```
yorkie, line 14:
```suggestion
export function pipeLog(readable?: NodeJS.ReadableStream, writable?: NodeJS.WritableStream, prefix?: string): void {
```

packages/costa/src/runtime.ts
```diff
@@ -14,9 +14,10 @@ import {
   readJson,
   writeJson
 } from 'fs-extra';
-import { download, constants, generateId, parsePluginName } from '@pipcook/pipcook-core';
+import { download, constants, generateId, parsePluginName, PluginStatus } from '@pipcook/pipcook-core';
 import tar from 'tar-stream';
 import { spawn, SpawnOptions } from 'child_process';
+import crypto from 'crypto';
```
yorkie, line 20: The builtin imports should be moved behind the third-party.

packages/costa/src/index.ts
```diff
@@ -105,6 +105,10 @@ export interface PluginPackage {
      * The plugin runtime
      */
     runtime?: 'nodejs' | 'python';
+    /**
+     * md5 of the tarball
+     */
+    md5?: string;
```
yorkie, line 111: Just not sure if this should be here. /cc @FeelyChau

packages/costa/src/runtime.ts
```diff
@@ -94,14 +95,20 @@ async function extractPackageJsonFromReadable(readable: NodeJS.ReadableStream, p
   return JSON.parse(packageJson);
 }
 
-function fetchPackageJsonFromGit(remote: string, head: string): Promise<any> {
+async function fetchPackageJsonFromGit(remote: string, head: string, source: PluginSource): Promise<any> {
```
yorkie, line 98: @FeelyChau @rickycao-qy Shall we use the template for these methods?

```js
const packageJson = new Package<Git | Npm | File | ...>(...);
packageJson.fetch(...);
```

packages/costa/src/runtime.ts
rickycao-qy: the file is the whole npm package and might be quite large?

packages/costa/package.json
rickycao-qy: done

packages/costa/src/runtime.ts
rickycao-qy: done

packages/costa/src/runtime.ts
rickycao-qy: done

packages/costa/src/utils.ts
rickycao-qy: done

packages/costa/src/runtime.ts
rickycao-qy: done

packages/costa/package.json
rickycao-qy: done

rickycao-qy:
This PR is hold for the reason that we would better to use some template methods for those functions to fetch from different sources. This to to be done

rickycao-qy:
Pipcook will change to v2.0 with different handling way of npm packages. Will close this PR temporarily. I will open it later if necessary

PR #606, opened 2020-10-09
bench: add config for circleCI

WenheLI:
This PR adds supports for circleci benchmarking scripts

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -7,8 +7,12 @@ jobs:
     docker:
         - image: cimg/node:14.11.0
     steps:
-      - checkout
-      - run: bash ./tools/benchmark.sh
+      - when:
+        condition:
+          euqal: [master, << pipeline.git.branch >>]
+        steps:
+          - checkout
+          - run: bash ./tools/benchmark.sh
 
 workflows:
   node-bench:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,0 +1,16 @@
+version: 2.1
+orbs:
+  node: circleci/node@3.0.0
+
+jobs:
+  bench:
+    docker:
+        - image: cimg/node:14.11.0
+    steps:
+      - checkout
+      - run: bash ./tools/benchmark.sh
+
+workflows:
+  node-bench:
+    jobs:
+      - bench
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -13,4 +13,4 @@ jobs:
 workflows:
   node-bench:
     jobs:
-      - bench
+      - benchmark
--- a/tools/benchmark.sh
+++ b/tools/benchmark.sh
@@ -42,7 +42,7 @@ t1=$(date +%s)
 t2=$(date +%s)
 mnist_time=$((t2-t1))
 
-if [ -z ${UPLOAD} ]
+if [ -z ${UPLOAD} ] || [ ${CIRCLE_BRANCH}!='master' ]
 then
   echo "{\"install_time\":${install_time}, \"init_time\":$init_time, \
   \"esbuild_time\":$esbuild_time, \"build_time\":$build_time, \"test_time\":$test_time, \
@@ -59,7 +59,7 @@ else
             --arg test_time $test_time \
             --arg time_stamp $(date +%s) \
             --arg mnist_time $mnist_time \
-            ". + [{install_time:$install_time, init_time:$init_time, \
+            ". + [{commitId: $CIRCLE_SHA1, install_time:$install_time, init_time:$init_time, \
                   esbuild_time:$esbuild_time, build_time:$build_time, test_time:$test_time, \
                   mnist_time:$mnist_time, timestamp: $time_stamp }]")  > data.json
 
--- a/tools/benchmark.sh
+++ b/tools/benchmark.sh
@@ -42,10 +42,14 @@ t1=$(date +%s)
 t2=$(date +%s)
 mnist_time=$((t2-t1))
 
-git clone https://github.com/imgcook/pipcook-benchmark.git
-cd pipcook-benchmark
+if [ -z ${UPLOAD} ]
+then
+  echo "{\"install_time\":${install_time}, \"init_time\":$init_time, \"esbuild_time\":$esbuild_time, \"build_time\":$build_time, \"test_time\":$test_time, \"mnist_time\":$mnist_time, \"timestamp\": $time_stamp }" | jq
+else 
+  git clone https://github.com/imgcook/pipcook-benchmark.git
+  cd pipcook-benchmark
 
-echo $(cat data.json | 
+  echo $(cat data.json | 
           jq --arg install_time $install_time \
             --arg esbuild_time $esbuild_time \
             --arg build_time $build_time \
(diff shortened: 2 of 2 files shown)


yorkie (changes requested):
Shall we run the task when PR updates?

yorkie:
And I think we can't run it on the current main branch: https://app.circleci.com/pipelines/github/alibaba/pipcook

WenheLI:
> https://app.circleci.com/pipelines/github/alibaba/pipcook

Set it only to run CI when PR is submitted.

tools/benchmark.sh
```diff
@@ -0,0 +1,49 @@
...
+            --arg esbuild_time $esbuild_time \
+            --arg build_time $build_time \
+            --arg init_time $init_time \
+            --arg test_time $test_time \
+            --arg time_stamp $(date +%s) \
+            --arg mnist_time $mnist_time \
+            '. + [{install_time:$install_time, init_time:$init_time, esbuild_time:$esbuild_time, build_time:$build_time, test_time:$test_time, mnist_time:$mnist_time, timestamp: $time_stamp }]')  > data.json
+
+git config user.email "eric.lwh@alibaba-inc.com"
+git config user.name "WenheLI"
```
yorkie, line 46: Could the above lines be configurable?

tools/benchmark.sh
```diff
@@ -0,0 +1,49 @@
...
+git add data.json
+git commit --allow-empty  -am"[circleci]: update data"
+git push -q https://${TOKEN}@github.com/imgcook/pipcook-benchmark.git
```
yorkie, line 49: I don't expect the result will be pushed to the benchmark repo if I run this locally, how about `sh tools/benchmark.sh --token=xxx`?

.circleci/config.yml
```diff
@@ -0,0 +1,16 @@
+version: 2.1
+orbs:
+  node: circleci/node@3.0.0
+
+jobs:
+  bench:
```
yorkie, line 6:
```suggestion
  benchmark:
```

yorkie:
We need documentation about this change.

tools/benchmark.sh
WenheLI: Make it configurable via CircleCI

tools/benchmark.sh
```diff
@@ -0,0 +1,67 @@
...
+            --arg build_time $build_time \
+            --arg init_time $init_time \
+            --arg test_time $test_time \
+            --arg time_stamp $(date +%s) \
+            --arg mnist_time $mnist_time \
+            '. + [{install_time:$install_time, init_time:$init_time, esbuild_time:$esbuild_time, build_time:$build_time, test_time:$test_time, mnist_time:$mnist_time, timestamp: $time_stamp }]')  > data.json
+
+if [ -z ${UPLOAD} ]
+then
+  echo "benchmark has been saved to data.json!"
```
yorkie, line 60: How about print the output only when not uploading?

tools/benchmark.sh
```diff
@@ -0,0 +1,67 @@
...
+else 
+  git config user.email ${EMAIL}
+  git config user.name ${USERNAME}
+  git add data.json
+  git commit --allow-empty  -am"[circleci]: update data"
+  git push -q https://${TOKEN}@github.com/imgcook/pipcook-benchmark.git
+fi
```
yorkie, line 67: EOF

tools/benchmark.sh
```diff
@@ -0,0 +1,67 @@
...
+test_time=$((t2-t1))
+
+t1=$(date +%s)
+./packages/cli/dist/bin/pipcook run ./example/pipelines/mnist-image-classification.json
+t2=$(date +%s)
+mnist_time=$((t2-t1))
+
+if [ -z ${UPLOAD} ]
+then
+  echo "{\"install_time\":${install_time}, \"init_time\":$init_time, \"esbuild_time\":$esbuild_time, \"build_time\":$build_time, \"test_time\":$test_time, \"mnist_time\":$mnist_time, \"timestamp\": $time_stamp }" | jq
```
yorkie, line 47: This line is too long.

tools/benchmark.sh
```diff
@@ -0,0 +1,67 @@
...
+
+  echo $(cat data.json | 
+          jq --arg install_time $install_time \
+            --arg esbuild_time $esbuild_time \
+            --arg build_time $build_time \
+            --arg init_time $init_time \
+            --arg test_time $test_time \
+            --arg time_stamp $(date +%s) \
+            --arg mnist_time $mnist_time \
+            '. + [{install_time:$install_time, init_time:$init_time, esbuild_time:$esbuild_time, build_time:$build_time, test_time:$test_time, mnist_time:$mnist_time, timestamp: $time_stamp }]')  > data.json
```
yorkie, line 60: Ditto.

tools/benchmark.sh
```diff
@@ -0,0 +1,67 @@
...
+
+  git config user.email ${EMAIL}
+  git config user.name ${USERNAME}
+  git add data.json
+  git commit --allow-empty  -am\"[circleci]: update data"
```
yorkie, line 65:
```suggestion
  git commit --allow-empty -am\"update data"
```

yorkie (changes requested):

.circleci/config.yml
```diff
@@ -0,0 +1,16 @@
...
+    docker:
+        - image: cimg/node:14.11.0
+    steps:
+      - checkout
+      - run: bash ./tools/benchmark.sh
+
+workflows:
+  node-bench:
+    jobs:
+      - bench
```
yorkie, line 16:
```suggestion
      - benchmark
```

yorkie:
@WenheLI What are the hooks to run this script in CircleCI?

WenheLI:
> @WenheLI What are the hooks to run this script in CircleCI?

This will run after every `PR` is merged.

yorkie:
> > @WenheLI What are the hooks to run this script in CircleCI?
> 
> This will run after every `PR` is merged.

It seems incorrect, we should run when PR updates.

WenheLI:
> > > @WenheLI What are the hooks to run this script in CircleCI?
> > 
> > 
> > This will run after every `PR` is merged.
> 
> It seems incorrect, we should run when PR updates.

Humm, that's true. I will update it later

WenheLI:
@yorkie Just came into my mind, since we want to keep recording the benchmark data, we should only run this benchmark when pr is merged. We can run another job for benchmarking on every commit.

yorkie:
Ok, then we should also declare the hooks explicitly that we used.

WenheLI:
> Ok, then we should also declare the hooks explicitly that we used.

Limit the CI only upload benchmark results when committing to master. In other cases, it will only print out the result.

yorkie:
> > Ok, then we should also declare the hooks explicitly that we used.
> 
> Limit the CI only upload benchmark results when committing to master. In other cases, it will only print out the result.

See https://circleci.com/docs/2.0/configuration-cookbook/#branch-filtering-for-job-steps, we need to config it in circle config.

yorkie (approved):

Issue #569, opened 2020-09-08
Data Collect Plugin re-use


rickycao-qy:
Pipeline works as following now:

1. provide a plugin-specific working dir for data-collect plugin. One data-collect plugin will download and maintain data in this working dir.

2. provide the same working dir to data-access plugin. Data-access plugin can also modify contents in this working directory.

This mechanism has the following problems that we hope data-collect plugin can make full use of this working directory. For example, they can reuse existing data, do some data cache, and optimize storage space. But since data-access plugin will also modify the same directory, the data-collect plugin will get lost

To solve this problem, I suggest we keep the current style for data-collect. Meanwhile, we provide a temp directory (job-specific) directory for data-access. This can be removed after the whole job is finished.

PR #607, opened 2020-10-10
pipeline, daemon: split data dir for data-collect and data-access

rickycao-qy:
Refer to Issue #569 

Pipeline works as following now:

provide a plugin-specific working dir for data-collect plugin. One data-collect plugin will download and maintain data in this working dir.

provide the same working dir to data-access plugin. Data-access plugin can also modify contents in this working directory.

This mechanism has the following problems that we hope data-collect plugin can make full use of this working directory. For example, they can reuse existing data, do some data cache, and optimize storage space. But since data-access plugin will also modify the same directory, the data-collect plugin will get lost

To solve this problem, I suggest we keep the current style for data-collect. Meanwhile, we provide a temp directory (job-specific) directory for data-access. This can be removed after the whole job is finished.

### DIFF
--- a/packages/costa/src/runnable.ts
+++ b/packages/costa/src/runnable.ts
@@ -99,8 +99,7 @@ export class PluginRunnable {
     const compPath = this.workingDir;
 
     debug(`make sure the component dir is existed.`);
-    await ensureDir(compPath + '/node_modules');
-    await ensureDir(this.dataDir);
+    await Promise.all([ ensureDir(compPath + '/node_modules'), ensureDir(this.dataDir) ]);
 
     debug(`bootstrap a new process for ${this.id}.`);
     this.handle = fork(__dirname + '/client/entry', [], {
--- a/packages/costa/src/runnable.ts
+++ b/packages/costa/src/runnable.ts
@@ -62,6 +62,11 @@ export class PluginRunnable {
   public workingDir: string;
 
   /**
+   * the current data directory for this runnable
+   */
+  public dataDir: string;
+
+  /**
    * The current state.
    */
   public state: 'init' | 'idle' | 'busy';
@@ -83,6 +88,7 @@ export class PluginRunnable {
     this.id = id || generateId();
     this.rt = rt;
     this.workingDir = path.join(this.rt.options.componentDir, this.id);
+    this.dataDir = path.join(this.workingDir, 'data');
     this.state = 'init';
     this.logger = logger || process;
   }
@@ -94,6 +100,7 @@ export class PluginRunnable {
 
     debug(`make sure the component dir is existed.`);
     await ensureDir(compPath + '/node_modules');
+    await ensureDir(this.dataDir);
 
     debug(`bootstrap a new process for ${this.id}.`);
     this.handle = fork(__dirname + '/client/entry', [], {
--- a/packages/daemon/src/runner/job-runner.ts
+++ b/packages/daemon/src/runner/job-runner.ts
@@ -244,7 +244,11 @@ export class JobRunner {
     const modelPath = path.join(this.opts.runnable.workingDir, 'model');
 
     await this.runDataCollect(dataDir, modelPath);
-    const dataset = await this.runDataAccess(dataDir);
+
+    // move plugin data directory data to job's data directory
+    await fs.copy(dataDir, this.opts.runnable.dataDir);
+
+    const dataset = await this.runDataAccess(this.opts.runnable.dataDir);
     await this.runDatasetProcess(dataset);
     await this.runDataProcess(dataset);
 
--- a/packages/daemon/src/runner/job-runner.ts
+++ b/packages/daemon/src/runner/job-runner.ts
@@ -8,6 +8,7 @@ import { PluginPackage, RunnableResponse, PluginRunnable } from '@pipcook/costa'
 import { PipelineEntity } from '../model/pipeline';
 import { JobParam, JobEntity } from '../model/job';
 import { Tracer, JobStatusChangeEvent } from '../service/trace-manager';
+import { copyDir } from '../utils';
 
 /**
  * plugin info from pipeline config
@@ -246,7 +247,7 @@ export class JobRunner {
     await this.runDataCollect(dataDir, modelPath);
 
     // move plugin data directory data to job's data directory
-    await fs.copy(dataDir, this.opts.runnable.dataDir);
+    await copyDir(dataDir, this.opts.runnable.dataDir);
 
     const dataset = await this.runDataAccess(this.opts.runnable.dataDir);
     await this.runDatasetProcess(dataset);
--- a/packages/daemon/test/runner/helper.test.ts
+++ b/packages/daemon/test/runner/helper.test.ts
@@ -146,6 +146,6 @@ describe('test the app service', () => {
     await assertNode.doesNotReject(async () => {
       await helper.copyDir(join(__dirname, 'src'), dest);
     }, 'copyDir should not be rejected');
-    assert.ok(symStub.calledOnce, 'symlink should be caled only once');
+    assert.ok(symStub.calledOnce, 'symlink should be called only once');
   });
 });
--- a/packages/daemon/test/runner/helper.test.ts
+++ b/packages/daemon/test/runner/helper.test.ts
@@ -127,4 +127,13 @@ describe('test the app service', () => {
     sinon.stub(fs, 'chmod');
     await helper.copyDir(src, dest);
   });
+  it('#should copy the symlink successfully', async () => {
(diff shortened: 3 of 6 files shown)


packages/daemon/src/service/pipeline.ts
```diff
@@ -210,9 +211,14 @@ export class PipelineService {
         dataDir
       }));
 
+      // create temp dir to use for data-access folder
+      tempDataDir = path.join(CoreConstants.PIPCOOK_TMPDIR, job.id);
```
yorkie, line 215: How about `path.join(runnable.workingDir, 'data')`?

rickycao-qy: done~

packages/daemon/src/service/pipeline.ts
```diff
@@ -214,14 +214,18 @@ export class PipelineService {
     // update the job status to running
     job.status = PipelineStatus.RUNNING;
     await JobModel.saveJob(job);
+    // temp data dir
+    const tempDataDir = path.join(runnable.workingDir, 'tmpdata');
+    await fs.ensureDir(tempDataDir);
```
yorkie, line 219: How about moving this to `JobRunner`, too?

packages/daemon/src/service/pipeline.ts
```diff
@@ -274,6 +278,9 @@ export class PipelineService {
       }
       throw err;
     } finally {
+      if (tempDataDir) {
+        await fs.remove(tempDataDir);
```
yorkie, line 282: Ditto, it could be removed at destroy.

packages/daemon/src/service/pipeline.ts
```diff
@@ -271,6 +271,7 @@ export class PipelineService {
       throw err;
     } finally {
       await runnable.destroy();
```
yorkie, line 273: Shall we move the `runnable.destroy()` to `JobRunner`, too?

packages/daemon/src/runner/job-runner.ts
```diff
@@ -57,8 +57,11 @@ export interface JobResult {
  */
 export class JobRunner {
   opts: RunnerOptions;
+  tmpdataDir: string;
+
   constructor(opts: RunnerOptions) {
     this.opts = opts;
+    this.tmpdataDir = path.join(this.opts.runnable.workingDir, 'tmpdata');
```
yorkie, line 64: The `JobRunner` should not operate on the file system, may I ask you to move this to costa's Runnable?

packages/daemon/src/runner/job-runner.ts
```diff
@@ -270,4 +278,8 @@ export class JobRunner {
       }
     };
   }
+
+  async destroy(): Promise<void> {
+    await fs.remove(this.tmpdataDir);
```
yorkie, line 283: By the way, Pipboard needs to load this data to analyze, thus we shouldn't remove it.

packages/daemon/src/runner/job-runner.ts
```diff
@@ -244,7 +247,12 @@ export class JobRunner {
     const modelPath = path.join(this.opts.runnable.workingDir, 'model');
 
     await this.runDataCollect(dataDir, modelPath);
-    const dataset = await this.runDataAccess(dataDir);
+
+    // copy data to a tmp directory
+    await fs.ensureDir(dataDir);
+    await fs.copy(dataDir, this.tmpdataDir);
```
yorkie, line 253:
```suggestion
    await fs.copy(dataDir, this.opts.runnable.dataDir);
```

It means we moving the pipeline data directory to a job's data directory.

rickycao-qy:
Have moved data dir operation to costa. Please have a look~

packages/daemon/src/runner/job-runner.ts
rickycao-qy: done

packages/daemon/src/runner/job-runner.ts
rickycao-qy: done

packages/daemon/src/runner/job-runner.ts
rickycao-qy: done

yorkie (approved):

FeelyChau (changes requested):

packages/daemon/src/runner/job-runner.ts
```diff
@@ -244,7 +244,11 @@ export class JobRunner {
     const modelPath = path.join(this.opts.runnable.workingDir, 'model');
 
     await this.runDataCollect(dataDir, modelPath);
-    const dataset = await this.runDataAccess(dataDir);
+
+    // move plugin data directory data to job's data directory
+    await fs.copy(dataDir, this.opts.runnable.dataDir);
```
FeelyChau, line 249: If use fs.copyFile with flag `fs.constants.COPYFILE_FICLONE`, this operation will be more faster.

FeelyChau (changes requested):

packages/daemon/src/runner/job-runner.ts
rickycao-qy: done

packages/costa/src/runnable.ts
```diff
@@ -94,6 +100,7 @@ export class PluginRunnable {
 
     debug(`make sure the component dir is existed.`);
     await ensureDir(compPath + '/node_modules');
+    await ensureDir(this.dataDir);
```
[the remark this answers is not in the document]
rickycao-qy: done

packages/daemon/test/utils/job-runner.test.ts
```diff
@@ -25,6 +26,10 @@ describe('test JobRunner', () => {
     sinon.restore();
   });
 
+  beforeEach(() => {
+    sinon.stub(util, 'copyDir').resolves();
```
FeelyChau, line 30: Not all the cases need to stub `copyDir`, I think we should stub it in the cases not in this hook.

FeelyChau (changes requested):

packages/daemon/test/utils/job-runner.test.ts
rickycao-qy: just remove this in the hook and add it to the place required

packages/daemon/test/runner/helper.test.ts
```diff
@@ -120,4 +121,34 @@ describe('test the app service', () => {
...
+  it('#should copy the symlink successfully', async () => {
+    const src = join(__dirname, 'src', 'sym.ts');
+    const dest = join(__dirname, 'dest');
+    try {
+      await fs.ensureSymlink(join(__dirname, 'helper.test.ts'), src);
+      const symStub = sinon.stub(fs, 'symlink');
+      await assertNode.doesNotReject(async () => {
+        await helper.copyDir(join(__dirname, 'src'), dest);
+      }, 'copyDir should not be rejected');
+      assert.ok(symStub.calledOnce, 'symlink should be caled only once');
```
FeelyChau, line 148:
```suggestion
      assert.ok(symStub.calledOnce, 'symlink should be called only once');
```

packages/daemon/test/runner/helper.test.ts
```diff
@@ -120,4 +121,34 @@ describe('test the app service', () => {
...
+  it('#should copy the folder successfully', async () => {
+    const src = join(__dirname, 'helper.test.ts');
+    const dest = join(__dirname, 'dest.ts');
+    const copyStub = sinon.stub(fs, 'copyFile');
+    const chmodStub = sinon.stub(fs, 'chmod');
+    try {
+      await assertNode.doesNotReject(async () => {
+        await helper.copyDir(src, dest);
+      }, 'copyDir should not be rejected');
+      assert.ok(copyStub.calledOnce, 'copyFile should be caled only once');
```
FeelyChau, line 133:
```suggestion
      assert.ok(copyStub.calledOnce, 'copyFile should be called only once');
```

packages/daemon/test/runner/helper.test.ts
```diff
@@ -120,4 +121,34 @@ describe('test the app service', () => {
...
+      assert.ok(chmodStub.calledOnce, 'chmod should be caled only once');
+    } finally {
+      await fs.remove(dest);
```
FeelyChau, line 136: We should mock the fs operation to keep unit test not side effect.

FeelyChau (changes requested):

packages/daemon/test/runner/helper.test.ts
rickycao-qy: done

FeelyChau (approved):
LGTM.

yorkie (approved):

Issue #585, opened 2020-09-10
To preserve accuracy in training set for datasets without test set


micheng-ZJU:
For datasets without test set, e.g., mock data, we would like to have an evaluation to tell the accuracy from training set. 
Both image classification models and object detection models.

yorkie:
@FeelyChau @rickycao-qy This requires we define the evaluation in more detail and save them in the database.

PR #608, opened 2020-10-10
core: add detail spec for evaluate result

rickycao-qy:
Refer to issue #585 

We need to support the evaluation result for both the train and test dataset. This PR is to support this case from the view of core spec. Meanwhile, we will change the plugins to support this feature.

### DIFF
--- a/packages/core/src/types/other.ts
+++ b/packages/core/src/types/other.ts
@@ -3,10 +3,19 @@ export interface Statistic {
   metricValue: number;
 }
 
-export interface EvaluateResult {
+export interface EvaluateResultItem {
   pass?: boolean;
   [key: string]: any;
 }
+
+export interface EvaluateResultOnDatasets {
+  train?: EvaluateResultItem;
+  validate?: EvaluateResultItem;
+  test?: EvaluateResultItem;
+  pass?: boolean;
+}
+
+export type EvaluateResult = EvaluateResultItem | EvaluateResultOnDatasets;
 
 export class EvaluateError extends TypeError {
   public result: EvaluateResult;


yorkie:
Shall we change the database as well?

packages/core/src/types/other.ts
```diff
@@ -3,11 +3,17 @@ export interface Statistic {
...
 }
 
-export interface EvaluateResult {
+export interface EvaluateResultItem {
   pass?: boolean;
   [key: string]: any;
 }
 
+export interface EvaluateResult {
+  trainResult?: EvaluateResultItem;
```
yorkie, line 12:
```suggestion
  train?: EvaluateResultItem;
```

The result might be duplicated with the interface name?

yorkie:
> Meanwhile, we will change the plugins to support this feature.

How does a plugin will write the train/test result?

rickycao-qy:
> > Meanwhile, we will change the plugins to support this feature.
> 
> How does a plugin will write the train/test result?

Have changed the type definition to support compatibility for current plugins. Later on, the plugin can be re-written to support new type definition

packages/core/src/types/other.ts
```diff
@@ -3,11 +3,19 @@ export interface Statistic {
   metricValue: number;
 }
 
-export interface EvaluateResult {
+export interface EvaluateResultItem {
   pass?: boolean;
   [key: string]: any;
 }
 
+export interface EvaluateResultDataset {
```
yorkie, line 11:
```suggestion
export interface EvaluateResultOnDatasets {
```

yorkie:
> the plugin can be re-written to support new type definition

How does the plugin to write the backward-compatible for this feature? For example, a new plugin returns `EvaluateResultOnDatasets` on an older Pipcook, shall we have a way to set up the plugin could be installed and ran for the current daemon?

yorkie:
And see https://coveralls.io/builds/34075298, the code coverage is decreased -1.2% at the pipeline.ts.

rickycao-qy:
> > the plugin can be re-written to support new type definition
> 
> How does the plugin to write the backward-compatible for this feature? For example, a new plugin returns `EvaluateResultOnDatasets` on an older Pipcook, shall we have a way to set up the plugin could be installed and ran for the current daemon?

To stop installing plugins seems not a good experience. I'm thinking if it's a good idea to specify `EvaluateResult = EvaluateResultItem & EvaluateResultDataset` since the old version just has `[key: string]: any`. In this case,  if a new plugin runs on the old daemon, it still conforms to this definition.

yorkie:
> To stop installing plugins seems not a good experience.

What I meant is just to verify if they can be installed on the host.

> I'm thinking if it's a good idea to specify EvaluateResult = EvaluateResultItem & EvaluateResultDataset since the old version just has [key: string]: any

Exactly, that's really the bad parts of using the dynamical definition before.

yorkie (approved):

WenheLI (approved):
LGTM

PR #609, opened 2020-10-10
Bug: Fix yarn lock

WenheLI:
This PR adds a temporary fix for the yarn lock issue. As we are using taobao mirror for downloading npm packages and the source addresses are logged by yarn, we will still download from the source. And this will slow down the downloading process while users are not in China.

As the yarn team won't fix it until yarn2. I added a quick fix to manually replace all taobao source to yarn.

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
   "scripts": {
     "build": "lerna run compile",
     "dev-build": "lerna run dev-compile --parallel --concurrency 10",
-    "postinstall": "node ./tools/init_dev.js -f && node ./tools/lock_treater.js",
+    "postinstall": "node ./tools/init_dev.js -f",
     "docsify": "npm run typedoc && http-server -c-1 ./docs",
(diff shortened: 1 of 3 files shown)


yorkie:
This looks too hacky, shall we have a fix on yarn?

yorkie:
Or we could update the yarn.lock only, and run `yarn config set registry https://registry.npm.taobao.org` locally.

WenheLI:
> `yarn config set registry https://registry.npm.taobao.org` locally.

IMO, The best practice should be updating the `yarn.lock` in CI. But this might result in a great conflict when developers are using different sources.

yorkie:
> > `yarn config set registry https://registry.npm.taobao.org` locally.
> 
> IMO, The best practice should be updating the `yarn.lock` in CI. But this might result in a great conflict when developers are using different sources.

It's not a good practice to update the lock by the script in this PR, may I ask you to update the lock file only?

yorkie (approved):

yorkie:
@WenheLI I removed the `bug` label because it's used for issue only.

PR #610, opened 2020-10-10
boa: don't set PYTHONHOME

yorkie:
As https://github.com/alibaba/pipcook/pull/582#issuecomment-706263289 pointed out, this causes spawning with incompatible python(like python2) won't work.


WenheLI:
The CI fails due to unresolved `$PYTHONHOME` seems like we still need to set it manually?

yorkie:
> The CI fails due to unresolved `$PYTHONHOME` seems like we still need to set it manually?

Yea, we have to manually remove PYTHONHOME when we want to spawn to use python2.

PR #618, opened 2020-10-20
daemon: support prefix id for cli operation

FeelyChau:
We can use prefix id for cli, like:
```sh
$ pipcook job list                                              
┌─────────┬────────────┬────────────┬───────────┬──────────────┬────────────────────────────┐
│ (index) │     id     │ pipelineId │  status   │ evaluatePass │         createdAt          │
├─────────┼────────────┼────────────┼───────────┼──────────────┼────────────────────────────┤
│    0    │ '5ck8g5no' │ 'he2tneg1' │ 'success' │     true     │ '2020-10-20T11:46:31.421Z' │
│    1    │ 'zwjlv9em' │ 'lx6k8m2h' │ 'success' │     true     │ '2020-10-20T11:45:44.670Z' │
│    2    │ '5ndv9z7x' │ '8mfffbnw' │ 'success' │     true     │ '2020-10-20T11:45:13.469Z' │
│    3    │ 'n54s87da' │ 'h6jpegv0' │ 'success' │     true     │ '2020-10-20T11:44:41.048Z' │
│    4    │ 'sv85mioz' │ 'pqekkfh7' │ 'success' │     true     │ '2020-10-20T11:43:38.121Z' │
│    5    │ 'azu0moag' │ 'ow97bsux' │ 'success' │     true     │ '2020-10-20T11:42:58.452Z' │
└─────────┴────────────┴────────────┴───────────┴──────────────┴────────────────────────────┘
$ pipcook job log 5
✖ multiple jobs found with prefix: 5
$ ipcook job log 5c      
ℹ ["Building prefix dict from the default dictionary ...
$ pipcook job log 5ck8g5no
ℹ ["Building prefix dict from the default dictionary ...
```

### DIFF
--- a/packages/daemon/src/app/controller/plugin.ts
+++ b/packages/daemon/src/app/controller/plugin.ts
@@ -1,7 +1,7 @@
 import { controller, inject, provide, get, post, del, put } from 'midway';
 import * as HttpStatus from 'http-status';
 import { BaseEventController } from './base';
-import { PluginManager } from '../../service/plugin';
+import { PluginManager, PluginEntity } from '../../service/plugin';
 import Debug from 'debug';
 const debug = Debug('daemon.app.plugin');
 
@@ -11,6 +11,22 @@ export class PluginController extends BaseEventController {
 
   @inject('pluginManager')
   pluginManager: PluginManager;
+
+  /**
+   * fetch plugin by plugin id prefix, return plugin entity object
+   * if zero or more than one plugin found, throw error
+   * @param prefix id prefix
+   */
+  async fetchPluginByIdPrefix(prefix: string): Promise<PluginEntity> {
+    const plugins = await this.pluginManager.findByPrefixId(prefix);
+    if (plugins.length > 1) {
+      return this.ctx.throw(HttpStatus.INTERNAL_SERVER_ERROR, `multiple plugins found with prefix: ${prefix}`);
+    }
+    if (plugins.length === 0) {
+      return this.ctx.throw(HttpStatus.NOT_FOUND, 'plugin not found');
+    }
+    return plugins[0];
+  }
 
   // TODO(feely): check if the plugin has been installed
   /**
@@ -36,17 +52,13 @@ export class PluginController extends BaseEventController {
   }
 
   /**
-   * delete plugin by name
+   * delete plugin by id
    */
   @del('/:id')
   public async remove() {
-    const plugin = await this.pluginManager.findById(this.ctx.params.id);
-    if (plugin) {
-      await this.pluginManager.uninstall(plugin);
-      this.ctx.success();
-    } else {
-      this.ctx.throw(HttpStatus.NOT_FOUND, `no plugin found by id ${this.ctx.params.id}`);
-    }
+    const plugin = await this.fetchPluginByIdPrefix(this.ctx.params.id);
+    await this.pluginManager.uninstall(plugin);
+    this.ctx.success();
   }
 
   /**
@@ -77,12 +89,8 @@ export class PluginController extends BaseEventController {
    */
   @get('/:id')
   public async get() {
-    const plugin = await this.pluginManager.findById(this.ctx.params.id);
-    if (plugin) {
-      this.ctx.success(plugin);
-    } else {
-      this.ctx.throw(HttpStatus.NOT_FOUND, 'no plugin found');
-    }
+    const plugin = await this.fetchPluginByIdPrefix(this.ctx.params.id);
+    this.ctx.success(plugin);
   }
 
   /**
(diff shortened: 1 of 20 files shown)


packages/daemon/src/model/plugin.ts
```diff
@@ -86,6 +84,7 @@ export class PluginModel extends Model {
   }
 
   static async findOrCreateByParams(creationParameter: CreationParameter): Promise<PluginEntity> {
+    console.log('create', creationParameter);
```
yorkie, line 87: Please remove the debug log.

packages/daemon/src/model/plugin.ts
```diff
@@ -49,24 +49,22 @@ export class PluginModel extends Model {
...
-
   static async findById(id: string): Promise<PluginEntity> {
     return (await PluginModel.findOne({ where: { id } }))?.toJSON() as PluginEntity;
   }
 
   static async findByIds(ids: string[]): Promise<PluginEntity[]> {
     return (await PluginModel.findAll({ where: { id: ids } })).map(plugin => plugin.toJSON() as PluginEntity);
   }
+
+  static async findByPrefixId(prefixId: string): Promise<PluginEntity[]> {
```
yorkie, line 60:
```suggestion
  static async searchById(keyword: string): Promise<PluginEntity[]> {
```

packages/daemon/src/model/plugin.ts
```diff
@@ -49,24 +49,22 @@ export class PluginModel extends Model {
     return (await PluginModel.findAll({ where })).map(plugin => plugin.toJSON() as PluginEntity);
   }
 
-  static async query(filter?: ListPluginsFilter): Promise<PluginEntity[]> {
```
yorkie, line 52: Why remove this method?

packages/daemon/src/model/pipeline.ts
```diff
@@ -59,6 +59,20 @@ export class PipelineModel extends Model {
     }))?.toJSON() as PipelineEntity;
   }
 
+  static async getPipelineByName(name: string): Promise<PipelineEntity> {
+    return (await PipelineModel.findOne({
+      where: { name }
+    }))?.toJSON() as PipelineEntity;
+  }
+
+  static async getPipelinesByPrefixId(prefixId: string): Promise<PipelineEntity[]> {
```
yorkie, line 68: Ditto, I don't get what PrefixId is from its name.

yorkie: In the name conventions, we will use "find**" if the result is an array.

packages/daemon/src/model/job.ts
```diff
@@ -42,6 +42,14 @@ export class JobModel extends Model {
     })).map(job => job.toJSON() as JobEntity);
   }
 
+  static async findByPrefixId(prefixId: string): Promise<JobEntity[]> {
```
yorkie, line 45: Ditto.

packages/daemon/src/app/controller/plugin.ts
```diff
@@ -77,23 +89,16 @@ export class PluginController extends BaseEventController {
    */
   @get('/:id')
   public async get() {
-    const plugin = await this.pluginManager.findById(this.ctx.params.id);
-    if (plugin) {
-      this.ctx.success(plugin);
-    } else {
-      this.ctx.throw(HttpStatus.NOT_FOUND, 'no plugin found');
-    }
+    const plugin = await this.fetchPluginByIdPrefix(this.ctx.params.id);
```
yorkie, line 92: I'm -1 on this change, we would keep the id to be the string that matches the actual id in the database, why not just implementing this in CLI?

yorkie (changes requested):

yorkie:
I'm -1 on this current implementation which changes the meaning of id in most APIs and it is incorrect.

packages/daemon/src/model/plugin.ts
FeelyChau: It's the same with function `list`.

FeelyChau:
> I'm -1 on this current implementation which changes the meaning of id in most APIs and it is incorrect.

How about add an option for fuzzy matching, default value is false, it could be compatible.

packages/daemon/src/model/plugin.ts
FeelyChau: It's a prefix of id, not keyword.

yorkie: I don't think we should introduce the prefix of id in daemon, which would add the complexity of our models.

yorkie:
> > I'm -1 on this current implementation which changes the meaning of id in most APIs and it is incorrect.
> 
> How about add an option for fuzzy matching, default value is false, it could be compatible.

Fuzzy matching doesn't work, because we have defined the `id` in APIs.

yorkie:
As I have commented on https://github.com/alibaba/pipcook/pull/618#discussion_r511729325, the daemon could provide a new API for searching the ids that match, and then the CLI use the result id to request the data.

FeelyChau:
> As I have commented on [#618 (comment)](https://github.com/alibaba/pipcook/pull/618#discussion_r511729325), the daemon could provide a new API for searching the ids that match, and then the CLI use the result id to request the data.

OK, I'll commit though another PR.

PR #622, opened 2020-10-28
boa: support TypeScript imports

yorkie:
This PR adds basic TypeScript support, the following is an example:

```ts
import * as boa from '@pipcook/boa';

const builtins = boa.builtins();
const mlist = builtins.list([1, 3, 5]);
console.log(JSON.stringify({ foobar: mlist }), '{"foobar":[1,3,5]}');

const colorsys = boa.import('colorsys');
const { len, min, max } = boa.builtins();
{
  const v = colorsys.rgb_to_hsv(0.2, 0.8, 0.4);
  console.log(v);
}
```

Note that the following import is not working because the `import` is a reversed keyword:

```ts
import { import } from '@pipcook/boa';
```


yorkie:
How about adding a compatible `pyimport` alias function?

```ts
import { pyimport } from '@pipcook/boa';
pyimport('numpy');
```

WenheLI:
> How about adding a compatible `pyimport` alias function?
> 
> ```ts
> import { pyimport } from '@pipcook/boa';
> pyimport('numpy');
> ```

`pyimport` should be fine.

yorkie:
> > How about adding a compatible `pyimport` alias function?
> > ```ts
> > import { pyimport } from '@pipcook/boa';
> > pyimport('numpy');
> > ```
> 
> `pyimport` should be fine.

Do you have a better name? Actually `pyimport` looks weird.

FeelyChau (approved):
LGTM.

PR #629, opened 2020-11-10
CI/CD: Add yarn timeout config to avoid npm install timeout          

WenheLI:
As title suggests, this pr is to avoid timeout while installing packages in github action

### DIFF
--- a/.yarnrc
+++ b/.yarnrc
@@ -1,2 +1,2 @@
 registry "https://registry.npmjs.com/"
-network-timeout 100000
+network-timeout 60000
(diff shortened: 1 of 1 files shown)


.yarnrc
```diff
@@ -1 +1,2 @@
 registry "https://registry.npmjs.com/"
+network-timeout 100000
```
yorkie, line 2: EOF

yorkie: Why do we use 100s?

WenheLI: Set it to 60s to align npm's default timeout limitation.
By default:
yarn : 30s
npm: 60s

Now, we config them to the same timeout

yorkie (approved):

PR #637, opened 2020-11-16
fix: merge coverage report

FeelyChau:


packages/daemon/package.json
```diff
@@ -52,7 +53,7 @@
     "clear": "midway-bin clean",
     "dev": "cross-env NODE_ENV=local midway-bin dev --port=6927 --ts",
     "debug": "cross-env NODE_ENV=local midway-bin debug --port=6927 --ts",
-    "cov": "midway-bin cov --ts",
+    "cov": "midway-bin cov --ts --nyc='-r lcov -r text-summary'",
```
yorkie, line 56: Why do we use text-summary, it causes us to lose some information. See https://istanbul.js.org/docs/advanced/alternative-reporters/#text-summary.

FeelyChau: `-r lcov -r text-summary` means nyc will generate file lcov.info on disk and text-summary on console.

packages/costa/.nycrc
```diff
@@ -22,5 +24,4 @@
   ],
   "all": true,
   "instrument": true
-}
-
+}
```
yorkie, line 27: EOF

packages/boa/package.json
```diff
@@ -10,7 +10,8 @@
     "test": "npm run test:js && npm run test:ts",
     "test:js": "tape ./tests/**/*.js",
     "test:ts": "ts-node ./node_modules/.bin/tape ./tests/**/*.ts",
-    "cov": "npx nyc --reporter=lcov --reporter=text-summary npm run test",
+    "cov": "nyc --reporter=text-summary npm run test",
+    "cov-report": "nyc report -r=lcov",
```
yorkie, line 14:
```suggestion
    "cov:report": "nyc report -r=lcov",
```

packages/core/package.json
```diff
@@ -13,7 +13,8 @@
     "clean": "rm -rf ./dist && rm -rf tsconfig.tsbuildinfo",
     "dev-compile": "tsc -b tsconfig.json",
     "compile": "tsc -b tsconfig.json",
-    "cov": "npx nyc --reporter=lcov --reporter=text-summary npm run test"
+    "cov": "nyc -r text-summary npm run test",
+    "cov-report": "nyc report -r lcov"
```
yorkie, line 17: Ditto

yorkie (changes requested):

packages/daemon/test/runner/helper.test.ts
```diff
@@ -4,6 +4,8 @@ import * as helper from '../../src/utils';
 import * as sinon from 'sinon';
 import * as fs from 'fs-extra';
 import { join } from 'path';
+import * as request from 'request-promise';
+const Bluebird = require( 'bluebird');
```
yorkie, line 8: Shall we use `Promise` instead?

yorkie (changes requested):

packages/daemon/test/runner/helper.test.ts
FeelyChau: Theoretically, `Bluebird` is compatible with `Promise`, but the data types are different, so we should mark the Promise object as any.
Updated.

yorkie: I can't get that why do we mark the `Promise` object as `any`?

FeelyChau: `Bulebird` is not completely the same with `Promise`, it will fail when check data type by typescript.

yorkie (approved):

PR #638, opened 2020-11-16
workflow: fix cache

FeelyChau:
1. We have refactored package manager to `yarn` from `npm`, so the lock file name changed from `package-lock.json` to `yarn.lock`, and only the root directory has the lock file.
2. The right way to use function `hashFiles` is: `hashFiles('fileA.json', 'fileB.json')`.


yorkie:
@FeelyChau May I ask you to write a bit about what to fix and how in this PR?

FeelyChau:
> @FeelyChau May I ask you to write a bit about what to fix and how in this PR?

OK, updated.

.github/workflows/build.yml
```diff
@@ -45,10 +45,10 @@ jobs:
           */*/node_modules
           packages/boa/pybind11
           packages/boa/pyodide
-        key: ${{ runner.os }}-${{ matrix.node_version }}-${{ hashFiles('**/package-lock.json,packages/boa/pybind11/Makefile,packages/boa/pyodide/Makefile') }}
+        key: ${{ runner.os }}-${{ matrix.node_version }}-yarn-${{ hashFiles('**/yarn.lock', 'packages/boa/pybind11/Makefile', 'packages/boa/pyodide/Makefile') }}
```
yorkie, line 48: Why update the key with appended yarn?

FeelyChau: YARN is a completely different package manager and the cache is not compatible.

WenheLI (approved):
LGTM

.github/workflows/build.yml
yorkie: But the file is different so that the key has been updated, though.

yorkie (approved):
Looks good to me

PR #642, opened 2020-11-17
costa: catch the error when subprocess send failed

FeelyChau:
We should set the callback when call `subprocess.send` for catching the asynchronous error.

### DIFF
--- a/packages/costa/src/runnable.ts
+++ b/packages/costa/src/runnable.ts
@@ -193,9 +193,14 @@ export class PluginRunnable {
    * @param op
    * @param msg
    */
-  private send(op: PluginOperator, msg?: PluginMessage): boolean {
+  private send(op: PluginOperator, msg?: PluginMessage): Promise<void> {
     const data = PluginProtocol.stringify(op, msg);
-    return this.handle.send(data);
+    return new Promise((resolve, reject) => {
+      const rst = this.handle.send(data, (err) => {
+        err ? reject(err) : resolve();
+      });
+      rst || reject(new Error('subprocess send failed'));
+    });
   }
   /**
    * Reads the message, it's blocking the async context util.
@@ -209,12 +214,11 @@ export class PluginRunnable {
   /**
    * Do send handshake message to runnable client, and wait for response.
    */
-  private async handshake(): Promise<boolean> {
-    await this.sendAndWait(PluginOperator.START, {
+  private async handshake(): Promise<PluginMessage> {
+    return this.sendAndWait(PluginOperator.START, {
       event: 'handshake',
       params: [ this.id ]
     });
-    return true;
   }
   /**
    * Wait for the next operator util receiving.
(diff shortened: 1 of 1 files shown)


packages/costa/src/runnable.ts
```diff
@@ -193,9 +193,14 @@ export class PluginRunnable {
    * @param op
    * @param msg
    */
-  private send(op: PluginOperator, msg?: PluginMessage): boolean {
+  private send(op: PluginOperator, msg?: PluginMessage): Promise<void> {
```
yorkie, line 196:
```suggestion
  private async send(op: PluginOperator, msg?: PluginMessage): Promise<void> {
```

packages/costa/src/runnable.ts
```diff
@@ -193,9 +193,14 @@ export class PluginRunnable {
...
     const data = PluginProtocol.stringify(op, msg);
-    return this.handle.send(data);
+    return new Promise((resolve, reject) => {
+      const rst = this.handle.send(data, (err) => {
+        err ? reject(err) : resolve();
```
yorkie, line 200:
```suggestion
        err ? reject(err) : resolve(rst);
```

packages/costa/src/runnable.ts
```diff
@@ -193,9 +193,14 @@ export class PluginRunnable {
...
+      });
+      rst || reject(new Error('subprocess send failed'));
```
yorkie, line 202: What does this mean?

packages/costa/src/runnable.ts
```diff
@@ -209,12 +214,11 @@ export class PluginRunnable {
   /**
    * Do send handshake message to runnable client, and wait for response.
    */
-  private async handshake(): Promise<boolean> {
-    await this.sendAndWait(PluginOperator.START, {
+  private async handshake(): Promise<PluginMessage> {
```
yorkie, line 217: Could you please keep this to return boolean?

packages/costa/src/runnable.ts
FeelyChau: `rst` is always true if `err` is null.

packages/costa/src/runnable.ts
FeelyChau: If `rst` is false, the callback will never been fired, so we need to reject an error by thin line.

packages/costa/src/runnable.ts
FeelyChau: I think the logic is clearer through exception handling.

packages/costa/src/runnable.ts
FeelyChau: Not necessary, no `await` used in thin function.

packages/costa/src/runnable.ts
FeelyChau: OK, I got u, Updated.

packages/costa/src/runnable.ts
```diff
@@ -193,9 +193,19 @@ export class PluginRunnable {
...
-    return this.handle.send(data);
+    return new Promise<void>((resolve, reject) => {
+      const success = this.handle.send(data, (err) => {
+        err ? reject(err) : resolve();
+      });
+      // if the message queue is full, the result will be false,
+      // and the callback will never been called,
+      // so we need to throw the error here
+      if (!success) {
+        reject(new Error('subprocess send failed'));
```
yorkie, line 206:
```suggestion
        reject(new TypeError('subprocess send failed'));
```

yorkie (approved):

PR #643, opened 2020-11-17
rfc: add new pipeline

yorkie:
See the markdown to review.

### DIFF
--- a/docs/rfcs/0000-new-pipeline.md
+++ b/docs/rfcs/0000-new-pipeline.md
@@ -249,3 +249,7 @@ And we MUST support pipeline check based on the JSON schema as:
 
 In 2.0, we are going to adapt to change it to the new pipeline syntax, however we still leave the 1.0
 pipeline parser.
+
+# Useful tools
+
+- JSON Schema: https://jsonschema.net/home
--- a/docs/rfcs/0000-new-pipeline.md
+++ b/docs/rfcs/0000-new-pipeline.md
@@ -1,0 +1,98 @@
+- Start Date: 2020-11-18
+- Target Major Version: 2.0
+- Reference Issues: (leave this empty)
+- Implementation PR: (leave this empty)
+- Author: @yorkie
+
+# Summary
+
+Introducing a new Pipeline syntax for supporting:
+
+- multiple nodes for the same plugin.
+- reinforcement learning
+- model deployment and distribution.
+
+# Basic example
+
+```json
+{
+  "version": "2.0",
+  "plugins": [
+    {
+      "type": "dataCollect",
+      "package": "@pipcook/plugins-chinese-poem-data-collect",
+      "params": {
+        "url": "foobar"
+      }
+    },
+    {
+      "type": "dataAccess",
+      "package": "@pipcook/plugins-textline-data-access"
+    },
+    {
+      "type": "dataProcess",
+      "package": "@pipcook/plugins-textline-data-process",
+      "params": {
+        "arg1": "foobar"
+      }
+    },
+    {
+      "type": "dataProcess",
+      "package": "@pipcook/plugins-textline-data-process",
+      "params": {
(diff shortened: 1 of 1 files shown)


yorkie:
I will add JSON schema later, please have reviews on other parts.

yorkie:
Added JSON schema definition, @WenheLI @rickycao-qy.

yorkie:
@rickycao-qy May I ask you to update the RL parts on this proposal?

docs/rfcs/0000-new-pipeline.md
```diff
@@ -0,0 +1,252 @@
...
+      "params": {
+        "epochs": 200
+      }
+    },
+    {
+      "type": "modelEvaluate",
+      "package": "@pipcook/plugins-tfjs-text-lstm-model-evaluate"
+    },
+    {
+      "type": "deploy",
```
rickycao-qy, line 61: Is it better to specify the type of 'modelDeploy', just to be in conformity with other plugins' types

docs/rfcs/0000-new-pipeline.md
```diff
@@ -0,0 +1,252 @@
...
+
+- multiple nodes for the same plugin.
+- model deployment and distribution.
+
+# Basic example
+
+```json
+{
+  "specVersion": "2.0",
+  "plugins": [
```
rickycao-qy, line 19: what if the user specifies more than one dataCollect/dataAccess/modelDeinfe plugins? Shall we make the whole plugins to be an array or just dataProcess to be an array?

docs/rfcs/0000-new-pipeline.md
yorkie: The name "modelDeploy" is the deprecated type, I use the new and common name is that what the plugin to deploy is not only the model, includes the runtime, data processing, and others.

docs/rfcs/0000-new-pipeline.md
yorkie: To make sure the extensibility of plugins, I think we could keep all plugins should be in the array, how about add additional rules to restrict the array such as limit only 1 "modelDefine" type?

rickycao-qy: This sounds good. We can modify the schema so that the numbers are limited to 1 for now.

yorkie: @rickycao-qy I have updated the schema below, the plugin items are split into multiple sub-schemas, which I defined the `maxItems` and `minItems` in different ones.

yorkie:
Within the separation of plugin items, I also defined some of the common parameters for every specific type of plugin, for example,

- added `url` to the params once the plugin type is "dataCollect"
- added `epochs` to the params once the plugin type is "modelTrain"
- ...

And we need to make sure these names are consistent with our core interfaces for plugin developers.

rickycao-qy:
> Within the separation of plugin items, I also defined some of the common parameters for every specific type of plugin, for example,
> 
> * added `url` to the params once the plugin type is "dataCollect"
> * added `epochs` to the params once the plugin type is "modelTrain"
> * ...
> 
> And we need to make sure these names are consistent with our core interfaces for plugin developers.

Just wondering if it's necessary to add these common parameters cuz the user may implement specific data-collect plugin just to collect specific data without url.

yorkie:
How about making the URL in `data-collect` be optional?

rickycao-qy:
> How about making the URL in `data-collect` be optional?

Sounds good to me

yorkie:
> > How about making the URL in `data-collect` be optional?
> 
> Sounds good to me

Updated :)

rickycao-qy (approved):
LGTM

FeelyChau (approved):
LGTM.

WenheLI (approved):
LGTM

PR #646, opened 2020-11-18
doc: fix the pipboard url

yorkie:
We have migrated the Pipboard from vercel to GitHub pages, this corrects the pipboard URI.


FeelyChau (approved):
LGTM.

Issue #631, opened 2020-11-11
pipeline: Pipeline is blocked on macos


FeelyChau:
![image](https://user-images.githubusercontent.com/10196611/98781245-00c77080-2431-11eb-94f0-509a088c7c02.png)

If run the mnist pipeline on mac, it will be blocked at:
```js
const tf = boa.import('tensorflow');
```

FeelyChau:
`Costa` runs plugins by 2 steps:
1. load plugin source by `require.resolve('path/to/plugin')`
2. call the entry function

In this case, the pipeline is blocked by step 1, so maybe we could add a monitor for job running, and the job running status could be `plugin-loading`, `plugin-running`, `failed`. If load timeout, the monitor could kill it and mark the job failed.

PR #647, opened 2020-11-18
costa: add the plugin not responding detection

yorkie:
This is able to tell the user if some plugin not responding on loading stage, and this is to implement the handler at https://github.com/alibaba/pipcook/issues/631#issuecomment-729560824.


yorkie:
This PR is to send a new WRITE message "plugin loaded" to notify if the plugin is loaded before a plugin is completed.

Note that, to send the message before the "pong", this PR introduces a new WRITE message "emit" which is used to send the message before "pong", and it adds the handlers reset when `sendAndWait()` is done.

yorkie:
The test case is WIP.

packages/costa/src/client/entry.ts
```diff
@@ -24,6 +24,18 @@ function recv(respOp: PluginOperator, ...params: string[]): void {
   }));
 }
 
+/**
+ * Send a `emit` message back from the client process, this will not stop the message handler.
+ * @param respOp the operator of response.
+ * @param msg
+ */
+function emit(respOp: PluginOperator, msg: string): void {
+  process.send(PluginProtocol.stringify(respOp, {
```
FeelyChau, line 33: Should catch the error from callback?

packages/costa/src/protocol.ts
```diff
@@ -20,7 +20,7 @@ export interface PluginResponse {
 export enum PluginOperator {
   START = 0x30,
   WRITE = 0x31,
-  READ = 0x32
+  READ = 0x32,
```
FeelyChau, line 23:
```suggestion
  READ = 0x32
```

packages/costa/src/runnable.test.ts
```diff
@@ -119,4 +121,18 @@ describe('start runnable in normal way', () => {
...
+
+  it('should destroy the runnable when it blocks on loading plugin', async () => {
+    const logger = process;
+    const costa = new CostaRuntime(opts);
+    const r2 = new PluginRunnable(costa, logger);
+    await r2.bootstrap({ logger, pluginNotRespondingTimeout: 500 });
+    const simple = await costa.fetch(path.join(__dirname, '../tests/plugins/nodejs-not-responding'));
+    await costa.install(simple, process);
+    try {
+      await r2.start(simple, { foobar: true });
```
FeelyChau, line 133: We could use `expectAsync` instead to make sure the asynchronous error has been caught.

packages/costa/src/runnable.ts
```diff
@@ -155,12 +169,36 @@ export class PluginRunnable {
...
       compPath + `/node_modules/${pkg.name}`);
 
-    const resp = await this.sendAndWait(PluginOperator.WRITE, {
-      event: 'start',
-      params: [
-        pkg,
-        ...args
-      ]
+    // wrap the PNR with the start logic.
+    const resp = await new Promise<PluginMessage>(async (resolve, reject) => {
```
FeelyChau, line 173: I think using `await` in `Promise` is not a good practice.

FeelyChau (changes requested):

packages/costa/src/runnable.ts
yorkie: Is there a better way this?

packages/costa/src/client/entry.ts
yorkie: There might not be a callback here :(

packages/costa/src/runnable.ts
```diff
@@ -244,22 +276,42 @@ export class PluginRunnable {
...
    * @param msg
    */
-  private async sendAndWait(op: PluginOperator, msg: PluginMessage): Promise<PluginMessage> {
-    await this.send(op, msg);
-    debug(`sent ${msg.event} for ${this.id}, and wait for response`);
+  private async sendAndWait(op: PluginOperator, msg: PluginMessage, timeout?: number): Promise<PluginMessage> {
+    return new Promise<PluginMessage>((resolve, reject) => {
+      let opNotRespondingTimer: NodeJS.Timer;
+      if (typeof timeout === 'number' && timeout > 0) {
+        opNotRespondingTimer = setTimeout(() => {
```
FeelyChau, line 283: I think we could insert the timer into function `send` like this:
```js
  /**
   * Reads the message, it's blocking the async context util.
   * @params timeout timeout value in millisecond
   */
  private async read(timeout = 0): Promise<PluginProtocol> {
     let timer;
      if (timeout > 0) {
        timer = setTimeout(() => {
          reject(new Error('read timeout'));
        }, timeout);
      }
      this.onread = (res: PluginProtocol) => {
        if (timer) {
          clearTimeout(timer);
        }
        resolve(res);
      };
      this.onreadfail = (err: Error) => {
        if (timer) {
          clearTimeout(timer);
        }
        reject(err);
      };
    });
  }
```
Is this a little more concise?

FeelyChau (changes requested):

yorkie:
@FeelyChau The CI is green now :)

FeelyChau (approved):

WenheLI (approved):
LGTM

PR #651, opened 2020-12-01
esbuild: ignore *.md

FeelyChau:
README.md is not source code file, so its changes should not trigger the build like `tsc-build`.


yorkie:
@FeelyChau please describe the reason for changing this.

FeelyChau:
> @FeelyChau please describe the reason for changing this.

Updated.

yorkie (approved):

PR #656, opened 2020-12-08
boa: enable python version switch in manual way

yorkie:
This PR tries to remove the specific version from the source code and gyp config, to switch the python version, change the function `getPythonVersion()` in tools/utils.js.

And I also add a TODO which is able to automatically detect the python library name from the corresponding conda source, which will be done in the next PR.


yorkie:
BTW, I will add documentation when the name detection is done, this PR is mainly to remove the hard versions.

FeelyChau (approved):
LGTM.

WenheLI (approved):
LGTM

PR #660, opened 2020-12-09
cli: check if BOOTSTRAP_HOME exists before starts

yorkie:
This PR checks for the BOOTSTRAP_HOME and prints a readable log instead of the stack trace.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -29,3 +29,5 @@ __pycache__
 *.pb
 coverage
 .nyc_output
+package-lock.json
+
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -29,3 +29,5 @@ __pycache__
 *.pb
 coverage
 .nyc_output
+package-lock.json
```
rickycao-qy, line 32: Is there a specific reason to ignore package-lock

yorkie: Because we have migrated to yarn, the package-lock should not be there anymore.

FeelyChau (approved):
LGTM.

WenheLI (approved):
LGTM

rickycao-qy (approved):
LGTM

PR #663, opened 2020-12-09
build: use new syntax to update env in workflow

yorkie:
See https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/ for more details.

It addresses this problem at https://github.com/alibaba/pipcook/pull/655#issuecomment-741675083.


FeelyChau (approved):
LGTM.

PR #666, opened 2020-12-11
daemon: Add tvm convertor

WenheLI:
This PR adds tvm convertor for our model output. 
Currently, it only supports keras format as this is our key use case.
Documentation, benchmark, and other formats support will be added in the coming commits.


packages/daemon/src/generator/boa.ts
```diff
@@ -0,0 +1,35 @@
+import { GenerateOptions } from '../service/pipeline';
```
yorkie, line 1: How about "generator/nodejs.ts"?

packages/daemon/src/generator/boa.ts
```diff
@@ -0,0 +1,35 @@
...
+import * as fs from 'fs-extra';
+import { JobEntity } from '../model/job';
+import * as path from 'path';
+
+export function boaGenerator(job: JobEntity, projPackage: any, dist: string, opts: GenerateOptions) {
```
yorkie, line 6: Shall we have a common interface for the generator function?

packages/daemon/src/generator/keras.ts
```diff
@@ -0,0 +1,74 @@
+import { exec } from 'child_process';
```
yorkie, line 1: Does this file seem to be a convertor than a generator?

packages/daemon/src/generator/keras.ts
```diff
@@ -0,0 +1,74 @@
...
+
+const boa = require('@pipcook/boa');
+
+export async function keras2Wasm(dist: string, projPackage: any, opts: GenerateOptions, inputLayer?: string) {
+  const relay = boa.import('tvm.relay');
+  const emcc = boa.import('tvm.contrib.emcc');
+  const keras = boa.import('tensorflow.keras');
+  const { dict, open, len } = boa.builtins();
+  exec(`wget http://ai-sample.oss-cn-hangzhou.aliyuncs.com/tvmjs/dist/tvmjs.bundle.js`, {cwd: dist});
+  exec(`wget http://ai-sample.oss-cn-hangzhou.aliyuncs.com/tvmjs/preload.js`, {cwd: dist});
```
yorkie, line 14: Why not using Node.js to download them?

yorkie: And we don't need to download them every time actually, and the above files should be versioned, too.

packages/daemon/src/generator/keras.ts
```diff
@@ -0,0 +1,74 @@
...
+
+  const model = keras.models.load_model(path.join(opts.modelPath, 'model.h5'));
+
+  const inputName = inputLayer ? inputLayer : 'input_1';
+  const inputShape = model.layers[0].input_shape[0];
+  if (len(inputShape) > 4) { return ; }
```
yorkie, line 20:
```suggestion
  if (len(inputShape) > 4) { return; }
```

packages/daemon/src/generator/keras.ts
```diff
@@ -0,0 +1,74 @@
+import { exec } from 'child_process';
+import * as path from 'path';
+import * as fs from 'fs-extra';
+import { GenerateOptions } from '../service/pipeline';
+
+const boa = require('@pipcook/boa');
+
+export async function keras2Wasm(dist: string, projPackage: any, opts: GenerateOptions, inputLayer?: string) {
```
yorkie, line 8: What's the return type?

packages/daemon/src/generator/keras.ts
```diff
@@ -0,0 +1,74 @@
...
+    this.imports = Module.wasmLibraryProvider.imports;
+    this.wasiImport = this.imports['wasi_snapshot_preview1'];
+  }
+  if (typeof module !== 'undefined' && module.exports) {
+    module.exports = EmccWASI;
+  }
+  `;
+
+  const result = templateHead + open(path.join(dist, 'model.wasi.js')).read() + templateTail;
+  const resultWriter = open(path.join(dist, 'model.wasi.js'), 'w');
```
yorkie, line 55: Ditto, the open is the blocking process, shall we use Node.js non-blocking fs functions instead?

packages/daemon/src/generator/keras.ts
```diff
@@ -0,0 +1,74 @@
...
+
+  const fileQueue = [];
+
+  const jsonPromise = fs.writeJSON(path.join(dist, 'modelSpec.json'), {
+    shape,
+    inputName
+  });
+  fileQueue.push(jsonPromise);
+  fileQueue.concat([
+    fs.copy(path.join(__dirname, `../../templates/wasm/predict.js`), `${dist}/index.js`),
```
yorkie, line 66:
```suggestion
    fs.copy(path.join(__dirname, '../../templates/wasm/predict.js'), `${dist}/index.js`),
```

packages/daemon/src/generator/tvmGeneratorParent.ts
```diff
@@ -0,0 +1,19 @@
+import { fork } from 'child_process';
+import { GenerateOptions } from '../service/pipeline';
+import * as path from 'path';
+import { existsSync } from 'fs-extra';
+
+export function tvmGenerator(dist: string, projPackage: any, opts: GenerateOptions) {
+  // only support keras for now; Other formats will be supported in the coming commits
+  if (! existsSync(path.join(opts.modelPath, 'model.h5'))) { return ; }
+  return new Promise((resolve, reject) => {
+    const client = fork(`${path.resolve(__dirname, 'tvmGenerator')}`, [JSON.stringify({
```
yorkie, line 10: Shall we have a processes pool that receives the converting request? The current implementation takes too many unnecessary costs.

packages/daemon/templates/wasm/predict.js
```diff
@@ -0,0 +1,41 @@
+const tvmjs = require("./tvmjs.bundle");
+const EmccWASI = require("./model.wasi");
+const fs = require('fs');
+const modelSpec = require("./modelSpec.json");
```
yorkie, line 4:
```suggestion
const modelSpec = require('./modelSpec.json');
```

packages/daemon/templates/wasm/predict.js
```diff
@@ -0,0 +1,41 @@
+const tvmjs = require("./tvmjs.bundle");
```
yorkie, line 1: Shall we use ESM?

packages/daemon/templates/wasm/predict.js
```diff
@@ -0,0 +1,41 @@
...
+const EmccWASI = require("./model.wasi");
+const fs = require('fs');
```
yorkie, line 3: This seems not working on the browser?

packages/daemon/templates/wasm/predict.js
yorkie: We need to make sure this file is working on our targets: browser/Node.js WASI :)

packages/daemon/templates/wasm/predict.js
WenheLI: IMO, a separate script is needed for browser env as we need to have file I/O in either case. And this part is not compatible.

yorkie: IMO it should be compatible because the developer is not expected to type the target when to start a pipeline.

WenheLI: True, I will try to figure out a way to make the I/O part compatible. The fallback solution is to generate two entry files: one for node, another one is for browser

packages/daemon/templates/wasm/predict.js
WenheLI: This file is browser/Node compatible. And it is generated based on Emcc WASI implementation.

packages/daemon/src/generator/keras.ts
WenheLI: Caching involves complex logic on version control and those files are relatively stable and small. For now, I think it is safe to download every time. We could fix it in another PR.

yorkie: There is no complexity to implement that, rename filenames with a specific version, and download them with a cache in memory.

packages/daemon/templates/wasm/predict.js
yorkie: What I mean is the file packages/daemon/templates/wasm/predict.js

packages/daemon/src/generator/boa.ts
WenheLI: Different generators might have their own parameter. I.E. tvm requires model type while boa does not

packages/daemon/src/generator/tvmGenerator.ts
```diff
@@ -0,0 +1,13 @@
+import { keras2Wasm } from '../convertor/keras';
+
+const validModel = ['keras'];
+
+const inputModel = process.argv[process.argv.length - 1];
+if (validModel.includes(inputModel)) {
+  const {dist, projPackage, opts, inputLayers = ''} = JSON.parse(process.argv[process.argv.length - 2]);
```
yorkie, line 7:
```suggestion
  const { dist, projPackage, opts, inputLayers = '' } = JSON.parse(process.argv[process.argv.length - 2]);
```

yorkie: We need to think about the failure on JSON.parse.

packages/daemon/src/generator/tvmGenerator.ts
```diff
@@ -0,0 +1,13 @@
...
+  if (inputModel === 'keras') {
+   keras2Wasm(dist, projPackage, opts, inputLayers);
```
yorkie, line 9:
```suggestion
    keras2Wasm(dist, projPackage, opts, inputLayers);
```

b.json
```diff
@@ -0,0 +1,3 @@
+{
+   'a': 1
```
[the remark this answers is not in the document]
WenheLI: The two are mis-commited. I will remove them.

packages/daemon/src/convertor/keras.ts
```diff
@@ -0,0 +1,75 @@
...
+  if (len(inputShape) > 4) { return; }
+  const shape = [1];
+  shape.push(inputShape[3]);
+  shape.push(inputShape[1]);
+  shape.push(inputShape[2]);
+
+  const [ mod, params ] = relay.frontend.from_keras(model, dict(boa.kwargs({[inputName]: shape})));
+  const [ graph, lib, param ] = relay.build(mod, boa.kwargs({
+    params,
+    target: 'llvm -mtriple=wasm32--unknown-emcc -system-lib'
```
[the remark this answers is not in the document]
WenheLI: This is under llvm context. I will try to find some description for it.
What I have done here is to let llvm to compile any platform(unknown) to emcc platform under wasm32.

packages/daemon/src/convertor/keras.ts
```diff
@@ -0,0 +1,75 @@
...
+  }));
+
+  lib.save(path.join(dist, 'model.bc'));
```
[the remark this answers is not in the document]
WenheLI: yep, it is under python and everything is sync.

packages/daemon/src/convertor/keras.ts
```diff
@@ -0,0 +1,75 @@
...
+  fileQueue.concat([
+    fs.copy(path.join(__dirname, '../../templates/wasm/predictBrowser.js'), `${dist}/predictBrowser.js`),
+    fs.copy(path.join(__dirname, '../../templates/wasm/predictNode.js'), `${dist}/predictNode.js`),
+    // write package.json
+    fs.outputJSON(dist + '/package.json', projPackage, { spaces: 2 }),
+  ]);
+
+  await Promise.all(fileQueue);
+
+  process.send('done');
```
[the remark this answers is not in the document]
WenheLI: We have to notify the parent thread that the convertor's task is done.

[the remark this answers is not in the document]
yorkie: Using signal code might be enough for this case.

packages/daemon/src/convertor/constant.ts
```diff
@@ -0,0 +1 @@
+export const TVM_PREFIX = 'http://ai-sample.oss-cn-hangzhou.aliyuncs.com/tvmjs';
```
yorkie, line 1:
```suggestion
export const TVM_BUNDLE_PREFIX = 'http://ai-sample.oss-cn-hangzhou.aliyuncs.com/tvmjs';
```

packages/daemon/src/convertor/keras.ts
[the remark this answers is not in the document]
yorkie: Is there another option with wasm32, like wasm64?

packages/daemon/src/generator/tvmGeneratorParent.ts
```diff
@@ -0,0 +1,19 @@
+import { fork } from 'child_process';
```
yorkie, line 1: @WenheLI As we have discussed, we should name this file as "generator/tvm.ts", and rename the `tvmGenerator` to another directory, or `generator/tvm.cli.ts`, so that we could get its relationship.

yorkie (changes requested):
@WenheLI Thanks for this excellent work! Added comments and two more suggestions:

1. `predictBrowser` and `predictNode` should be merged, we could make it be compatible with Node.js and browser by feature/environment detection.
2. most of this change looks great to me, thus it's time to add the unit tests and the manual.

packages/daemon/src/convertor/keras.ts
```diff
@@ -0,0 +1,75 @@
+import * as path from 'path';
+import * as fs from 'fs-extra';
+import { GenerateOptions } from '../service/pipeline';
+import { download } from '@pipcook/pipcook-core/dist/utils/public';
+
+const boa = require('@pipcook/boa');
```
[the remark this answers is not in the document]
WenheLI: BOA's syntax violates tsc's checking, we should re-consider using ES module within this project.

packages/daemon/templates/wasm/predict.js
```diff
@@ -0,0 +1,96 @@
+if (typeof process !== 'undefined') {
+  // running under node
+  const tvmjs = require('./tvmjs.bundle');
+  const EmccWASI = require("./model.wasi");
+  const fs = require('fs-extra');
+  const modelSpec = require('./modelSpec.json');
+  const graph = require('./modelDesc.json');
+  
+  const loadModel = async () => {
```
yorkie: How about making this function export, too?

packages/daemon/templates/wasm/predict.js
```diff
@@ -0,0 +1,96 @@
...
+      const wasmSource = fs.readFile('./model.wasi.wasm');
+      const paramsSource = fs.readFile('./modelParams.parmas');
+      Promise.all([wasmSource, paramsSource]).then(async ([wasm, params]) => {
+        const tvm = await tvmjs.instantiate(wasm, new EmccWASI());
+        const param = new Uint8Array(params);
+        const ctx = tvm.cpu(0);
+        const sysLib = tvm.systemLib();
+        model = tvm.createGraphRuntime(JSON.stringify(graph), sysLib, ctx);
+        model.loadParams(param);
+        resolve({model, tvm, ctx});
```
yorkie: In a thenable function, returning value is enough.

packages/daemon/templates/wasm/node.js
```diff
@@ -0,0 +1,43 @@
...
+
+const predict = async (input) => {
+  if (!model) {
+    const rets = await loadModel();
+    model = rets.model;
+    tvm = rets.tvm;
+    ctx = rets.ctx;
+  }
+
+  const inputData = tvm.empty(modelSpec.shape, "float32", tvm.cpu());
```
yorkie, line 33:
```suggestion
  const inputData = tvm.empty(modelSpec.shape, 'float32', tvm.cpu());
```

packages/daemon/templates/wasm/node.js
```diff
@@ -0,0 +1,43 @@
...
+  const output = model.getOutput(0);
+  inputData.copyFrom(input);
+  model.setInput(modelSpec.inputName, inputData);
+  model.run();
+  await ctx.sync();
+  console.log(output.toArray())
```
yorkie, line 39:
```suggestion
  console.log(output.toArray());
```

docs/manual/intro-to-pipeline.md
```diff
@@ -51,15 +51,18 @@ $ pipcook run /path/to/your/pipeline-config.json
...
-   ┗ 📜index.js
+output
+├─ logs
+├─ nodejs
+├─ package.json
+└─ wasm
 ```
 
-To get started with your trained model, follow the below steps:
+There are two entry-points here, one for nodejs powered by boa; another one is used for wasm (WebAssembly), a universial virtual machine, powered by TVM.
```
yorkie, line 61:
```suggestion
There are two entry-points here, one for Node.js powered by boa; another one is used for wasm (WebAssembly), a universial virtual machine, powered by TVM.
```

docs/manual/intro-to-pipeline.md
```diff
@@ -51,15 +51,18 @@ $ pipcook run /path/to/your/pipeline-config.json
...
-📂output
-   ┣ 📂logs
-   ┣ 📂model
-   ┣ 📜package.json
-   ┣ 📜metadata.json
-   ┗ 📜index.js
+output
+├─ logs
+├─ nodejs
+├─ package.json
```
yorkie, line 57: To be removed :)

yarn.lock
```diff
@@ -1379,6 +1379,14 @@
   dependencies:
     "@types/node" ">= 8"
 
+"@pipcook/plugins-bayesian-model-define@1.1.0":
```
yorkie, line 1382: It seems we shouldn't depend on any plugin.

yorkie (changes requested):

packages/cli/src/actions/add.ts
```diff
@@ -0,0 +1,24 @@
+import { PIPCOOK_BOA } from '../constants';
+import { execAsync, logger } from '../utils/common';
+
+const supportDeps = [ 'tensorflow', 'tvm' ];
```
yorkie, line 4: How about adding a full option?

packages/cli/src/actions/add.ts
```diff
@@ -0,0 +1,24 @@
...
+
+/**
+ * Add missing deps
+ */
+const add = async () => {
+  if (process.argv.length !== 4) { return; }
+  const name = process.argv[3];
+  if (supportDeps.includes(name)) {
+    if (name === 'tvm') {
+      logger.info('Installing tvm ...');
```
yorkie, line 14: This might be unnecessary because the below command would also print the related installation log.

packages/cli/src/actions/add.ts
```diff
@@ -0,0 +1,24 @@
...
+      await execAsync(`node ${PIPCOOK_BOA}/tools/bip.js install tlcpack -f https://tlcpack.ai/wheels`);
+    } else {
+      await execAsync(`node ${PIPCOOK_BOA}/tools/bip.js install ${name}`);
+    }
+  } else {
+    console.warn(`You are trying install ${name}, while it is not supported at this time`);
```
yorkie, line 20: We need to give the possible acceptable values instead of telling the input is incorrect :)

packages/cli/src/bin/pipcook.ts
```diff
@@ -87,6 +88,10 @@ import devPlugin from '../actions/dev-plugin';
     .description('initialize plugin development environment')
     .action(devPlugin);
 
+  program
+    .command('add')
```
yorkie, line 92: add a description as well :)

packages/daemon/src/convertor/keras.ts
```diff
@@ -0,0 +1,97 @@
...
+const { dict, len, open } = boa.builtins();
+let relay = null;
+let emcc = null;
+let keras = null;
+
+try {
+  relay = boa.import('tvm.relay');
+} catch {
+  console.error('Does not detect tvm environment here; Please try to use \n pipcook add tvm \n To fix it');
+  throw new Error('tvm does not install');
```
yorkie, line 17:
```suggestion
  throw new TypeError('tvm is required, run "pipcook add tvm".');
```

packages/daemon/src/convertor/keras.ts
```diff
@@ -0,0 +1,97 @@
...
+}
+
+try {
+  emcc = boa.import('tvm.contrib.emcc');
+} catch {
+  console.error('Does not detect emscripten environment here; Please try to install emscripten');
+  throw new Error('emscripten does not install');
```
yorkie, line 24: Ditto

packages/daemon/src/convertor/keras.ts
```diff
@@ -0,0 +1,97 @@
...
+}
+
+try {
+  keras = boa.import('tensorflow.keras');
+} catch {
+  console.error('Does not detect TensorFlow environment here; Please try to use \n pipcook add tensorflow \n To fix it');
+  throw new Error('Tensorflow does not install');
+}
```
yorkie, line 32: Shall we add a util function for these duplicated blocks?

packages/daemon/src/convertor/keras.ts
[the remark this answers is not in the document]
yorkie: Ping @WenheLI

packages/daemon/templates/wasm/browser.js
```diff
@@ -0,0 +1,43 @@
+import * as tvmjs from './tvmjs.bundle';
+import * as EmccWASI from './model.wasi';
+
+const loadModel = async () => {
+  const reqList = [fetch('./modelSpec.json'), fetch('./modelDesc.json'), fetch('./model.wasi.wasm'), fetch('./modelParams.parmas')];
+  const [spec, desc, wasmSource, params] = await Promise.all(reqList);
```
yorkie, line 6:
```suggestion
  const [ spec, desc, wasmSource, params ] = await Promise.all(reqList);
```

packages/daemon/templates/wasm/browser.js
```diff
@@ -0,0 +1,43 @@
+import * as tvmjs from './tvmjs.bundle';
+import * as EmccWASI from './model.wasi';
+
+const loadModel = async () => {
+  const reqList = [fetch('./modelSpec.json'), fetch('./modelDesc.json'), fetch('./model.wasi.wasm'), fetch('./modelParams.parmas')];
```
yorkie, line 5: In multiple lines?

packages/daemon/templates/wasm/browser.js
```diff
@@ -0,0 +1,43 @@
...
+  const [spec, desc, wasmSource, params] = await Promise.all(reqList);
+  const wasm = await wasmSource.arrayBuffer();
+  const param = await params.arrayBuffer();
```
yorkie, line 8: What are the differences between `params` with `param`?

packages/daemon/templates/wasm/browser.js
```diff
@@ -0,0 +1,43 @@
...
+  const graph = await desc.json();
+  const modelSpec = await spec.json();
+  const tvm = await tvmjs.instantiate(wasm, new EmccWASI());
+  const param = new Uint8Array(param);
+  const ctx = tvm.cpu(0);
+  const sysLib = tvm.systemLib();
+  model = tvm.createGraphRuntime(JSON.stringify(graph), sysLib, ctx);
+  model.loadParams(param);
+  return {model, tvm, ctx, modelSpec};
```
yorkie, line 17:
```suggestion
  return { model, tvm, ctx, modelSpec };
```

packages/daemon/templates/wasm/browser.js
```diff
@@ -0,0 +1,43 @@
...
+}
+
+let model, tvm, ctx, modelSpec;
+
+const predict = async (input) => {
```
yorkie, line 22: Why do we use the arrow function here, it seems not access to the `this`.

yorkie:
@WenheLI Please fix the lint errors and update Chinese documentation, too :)

packages/cli/src/actions/add.ts
```diff
@@ -0,0 +1,24 @@
+import { PIPCOOK_BOA } from '../constants';
+import { execAsync, logger } from '../utils/common';
+
+const supportDeps = [ 'tensorflow', 'tvm' ];
+
+/**
+ * Add missing deps
+ */
+const add = async () => {
```
yorkie, line 9: Oops, we should implement this function on the server-side.

WenheLI:
Too many conflicts, reopen another one https://github.com/alibaba/pipcook/pull/691

PR #675, opened 2020-12-19
notebook: upgrade the nodejs version to 12.19

yorkie:
As we have upgraded to the v1.3.0, it validates the Node.js version to be bigger than 10.19.


rickycao-qy (approved):
LGTM

PR #694, opened 2021-01-20
build: make boa work with Node10 and NAPI3

yorkie:
Node.js 12 requires a higher GLIBC, which won't be supported on some platforms, thus this PR downgrades the Node.js and NAPI requirements. Now the boa is able to work with Node.js 10.13 and NAPI3.

### DIFF
--- a/packages/boa/lib/worker.js
+++ b/packages/boa/lib/worker.js
@@ -8,9 +8,15 @@ const { PythonObject, NODE_PYTHON_HANDLE_NAME } = require('bindings')('boa');
 const { wrap } = require('./proxy');
 
 const TYPE_ID = 'SHARED_PYTHON_OBJECT_TYPE';
-if (process.moduleLoadList.indexOf('NativeModule worker_threads') > 0) {
-  ({ isMainThread, workerData } = require('worker_threads'));
+try {
+  require.resolve('worker_threads');
   supportWorkerThreads = true;
+} catch (err) {
+  supportWorkerThreads = false;
+}
+
+if (supportWorkerThreads === true) {
+  ({ isMainThread, workerData } = require('worker_threads'));
 }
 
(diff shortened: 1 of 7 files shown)


yorkie:
@FeelyChau The CI is green :)

packages/boa/package.json
```diff
@@ -53,7 +53,8 @@
     "eslint": "^7.15.0",
     "nyc": "^15.1.0",
     "tape": "^5.0.1",
-    "ts-node": "^8.6.2"
+    "ts-node": "^8.6.2",
+    "typescript": "^4.1.3"
```
FeelyChau, line 57: Why import the typescript?

packages/boa/lib/worker.js
```diff
@@ -1,19 +1,35 @@
...
+let isMainThread = false;
+let workerData = null;
+let supportWorkerThreads = false;
+
 const { PythonObject, NODE_PYTHON_HANDLE_NAME } = require('bindings')('boa');
 const { wrap } = require('./proxy');
 
 const TYPE_ID = 'SHARED_PYTHON_OBJECT_TYPE';
+try {
+  require.resolve('worker_threads');
```
FeelyChau, line 12: Could we do the flag though the version of process instead trying?

FeelyChau (changes requested):

packages/boa/lib/worker.js
yorkie: I think a feature-based check is better than a version-based.

packages/boa/package.json
yorkie: Oops.

FeelyChau (approved):
LGTM.

rickycao-qy (approved):
LGTM

PR #702, opened 2021-01-24
core: migrate jasmine to ava

FeelyChau:
1. migrate to ava from jasmine
2. remove download progress when call the download api, it should be implemented by cli/daemon/plugin.
3. remove api `getOsInfo`, we can use `platform()` instead of it.
4. remove unused dependency: decompress-zip, rxjs.


yorkie (approved):

WenheLI (approved):
LGTM

PR #708, opened 2021-01-25
boa: migrate tests to the ava framework

yorkie:


packages/boa/tests/base/basic.js
```diff
@@ -1,52 +1,48 @@
...
+  t.throws(() => boa.kwargs(false), { instanceOf: TypeError });
+  t.throws(() => boa.kwargs(true), { instanceOf: TypeError });
+  t.throws(() => boa.kwargs('foobar'), { instanceOf: TypeError });
+  t.throws(() => boa.kwargs(123), { instanceOf: TypeError });
 });
 
 test('hash function', t => {
-  t.ok(builtins.__hash__());
-  t.equal(builtins['__notexists__'], undefined);
+  t.is(Boolean(builtins.__hash__()), true);
```
FeelyChau, line 14:
```suggestion
  t.true(Boolean(builtins.__hash__()));
```

packages/boa/tests/base/delegator-builtins-generator.js
```diff
@@ -22,11 +21,10 @@ test('builtins.generator with next', t => {
 
   let val = generator.next();
 
-  t.equal(val.value, count);
-  t.equal(val.done, false);
+  t.is(val.value, count);
+  t.is(val.done, false);
```
FeelyChau, line 25:
```suggestion
  t.false(val.done);
```

packages/boa/tests/numpy/routines.matrix.js
```diff
@@ -9,48 +9,42 @@ boa.import('numpy.matlib');
 test('numpy.mat: interpret the input as a matrix.', t => {
   const x = np.array([[1, 2], [3, 4]]);
   const m = np.asmatrix(x);
-  t.equal(JSON.stringify(m.tolist()), '[[1,2],[3,4]]');
-  t.equal(type(m).__name__, 'matrix');
-  t.end();
+  t.is(JSON.stringify(m.tolist()), '[[1,2],[3,4]]');
```
FeelyChau, line 12: If `m.tolist()` return an array, we can use `deepEqual` like:
```js
t.deepEqual(m.tolist(), [[1,2],[3,4]]);
```

packages/boa/tests/numpy/routines.matrix.js
```diff
@@ -94,22 +88,22 @@ test('numpy.matlib.randn: return a random matrix with data from the “standard
...
+  t.is(`${m1.A1}`, '[21 22 23 24]');
   // simple test
   console.log(m1.sub(10));
   console.log(m1.mul(10));
   console.log(m1.rmul(3));
   console.log(m1.imul(3));
   console.log(m1.pow(2));
   console.log(m1.ipow(3));
-  t.end();
+  t.pass();
```
FeelyChau, line 106: Could the result be checked like `add`?

packages/boa/tests/numpy/ufunc.js
```diff
@@ -1,8 +1,8 @@
-const test = require('tape');
+const test = require('ava');
 const boa = require('../../');
 const np = boa.import('numpy');
 
 test('the ndarry constructor', t => {
   console.log(`${np.typecodes['All']}`);
-  t.end();
+  t.pass();
 });
```
FeelyChau, line 8: Missing EOF.

packages/boa/tests/stdlib/builtins.js
```diff
@@ -1,73 +1,64 @@
...
-  t.strictEqual(builtins.False, false);
-  t.strictEqual(builtins.None, null);
-  t.ok(builtins.NotImplemented);
-  t.ok(builtins.Ellipsis);
-  t.ok(builtins.__debug__);
-  t.end();
+  t.is(builtins.True, true);
+  t.is(builtins.False, false);
+  t.is(builtins.None, null);
+  t.is(Boolean(builtins.NotImplemented), true);
```
FeelyChau, line 9: t.true()

FeelyChau (changes requested):

packages/boa/tests/base/basic.js
yorkie: true is the keyword, I'd prefer `is` instead of the `true` method.

packages/boa/tests/base/delegator-builtins-generator.js
yorkie: Ditto

packages/boa/tests/numpy/routines.matrix.js
yorkie: tolist() won't return a normal JS array actually, that's why we used the string comparison here. And this PR is to migrate the framework only :(

packages/boa/tests/numpy/routines.matrix.js
yorkie: We need to verify if this function could be called here, because the computation logic is implemented in CPython.

packages/boa/tests/stdlib/builtins.js
yorkie: Ditto, see https://github.com/alibaba/pipcook/pull/708#discussion_r564291750.

FeelyChau (approved):
LGTM.

PR #712, opened 2021-02-07
fix copyFile block with option COPYFILE_FICLONE

FeelyChau:
Fix: `fs.copyFile(src, dest, fs.constants.COPYFILE_FICLONE)`  blocks the promise train, because `grancful-fs` which is depended by `fs-extra` misses the callback function. see:
https://github.com/isaacs/node-graceful-fs/pull/202


WenheLI (approved):
LGTM

PR #721, opened 2021-03-17
pipcook2.0: Support training data in a random way

Ahkari:
Support training data in a random way.


packages/core/src/types/runtime.ts
```diff
@@ -64,6 +64,15 @@ export interface DataAccessor<T = DefaultType> {
...
   seek: (pos: number) => Promise<void>;
 }
+export interface RandomDataAccessor<T = DefaultType> extends DataAccessor {
+  randomMap: Array<number>;
+  dataLength: number;
+  init: (SourceMeta: ImageDataSourceMeta) => void;
+  nextBatch: (batchSize: number) => Promise<Array<Sample<T>> | null>;
+  nextBatchRandom: (batchSize: number) => Promise<Array<Sample<T>> | null>;
+  resetRandom: (randomSeed: Array<number> | null) => Promise<Array<number>>;
+  createRandom: (dataLength: number) => Array<number>;
```
FeelyChau, line 74: Ditto.

packages/core/src/types/runtime.ts
```diff
@@ -64,6 +64,15 @@ export interface DataAccessor<T = DefaultType> {
   nextBatch: (batchSize: number) => Promise<Array<Sample<T>> | null>;
   seek: (pos: number) => Promise<void>;
 }
+export interface RandomDataAccessor<T = DefaultType> extends DataAccessor {
+  randomMap: Array<number>;
+  dataLength: number;
+  init: (SourceMeta: ImageDataSourceMeta) => void;
```
FeelyChau, line 70: They are not a part of the api, but the implement, I think it's better to move them into class.

packages/core/src/types/runtime.ts
```diff
@@ -64,6 +64,15 @@ export interface DataAccessor<T = DefaultType> {
...
+  nextBatch: (batchSize: number) => Promise<Array<Sample<T>> | null>;
```
FeelyChau, line 71: `nextBatch` has been defined in `DataAccessor`, do you mean `next`?

packages/core/src/types/runtime.ts
```diff
@@ -72,6 +81,13 @@ export interface DataSourceApi<T = DefaultType> {
   evaluate?: DataAccessor<T>;
 }
 
+export interface RandomDataSourceApi<T = DefaultType> {
+  getDataSourceMeta: () => Promise<DataSourceMeta>;
+  test: DataAccessor<T>;
+  train: DataAccessor<T>;
+  evaluate?: DataAccessor<T>;
```
FeelyChau, line 88:
```suggestion
  test: RandomDataAccessor<T>;
  train: RandomDataAccessor<T>;
  evaluate?: RandomDataAccessor<T>;
```

packages/core/src/types/runtime.ts
```diff
@@ -64,6 +64,15 @@ export interface DataAccessor<T = DefaultType> {
   nextBatch: (batchSize: number) => Promise<Array<Sample<T>> | null>;
   seek: (pos: number) => Promise<void>;
 }
+export interface RandomDataAccessor<T = DefaultType> extends DataAccessor {
```
FeelyChau, line 67: `RandomDataAccessor` is a little confusing. It extends the `DataAccessor` and realizes random reading, but it is the real **data accessor** for the scripts, so I suggest rename it to `DataAccessor` and rename origin `DataAccessor` to `SequentialDataAccessor`， @WenheLI @rickycao-qy @yorkie How do you think?

FeelyChau (changes requested):

packages/cli/src/standalone-impl.ts
```diff
@@ -13,32 +15,90 @@ import {
...
+    while (batchSize) {
+      buffer.push(this.nextRandom());
+      batchSize--;
+    }
+    return Promise.all(buffer);
+  }
   seek(pos: number): Promise<void> {
     return this.dataAccessor.seek(pos);
   }
+  createRandom(dataLength = 0): Array<number> {
```
FeelyChau, line 55:
```suggestion
  createRandom(dataLength: number): number {
```

packages/cli/src/standalone-impl.ts
```diff
@@ -13,32 +15,90 @@ import {
...
+  }
   next(): Promise<Sample<T> | null> {
     return this.dataAccessor.next();
   }
   nextBatch(batchSize: number): Promise<Array<Sample<T>> | null> {
     return this.dataAccessor.nextBatch(batchSize);
   }
+  nextRandom(): Promise<Sample<T> | null> {
+    if ( this.randomIndex >= this.dataLength ) {
+      this.dataAccessor.seek(this.randomIndex++);
```
FeelyChau, line 38:
```suggestion
      return null;
```

FeelyChau (changes requested):

packages/cli/src/standalone-impl.ts
Ahkari: I want to simulate the behavior like Function next().

packages/cli/src/standalone-impl.ts
```diff
@@ -13,32 +15,90 @@ import {
...
+  }
   seek(pos: number): Promise<void> {
     return this.dataAccessor.seek(pos);
   }
+  createRandom(dataLength = 0): Array<number> {
+    if (dataLength === 0)
+      return [];
+    let i = 0;
+    let index = 0;
+    let temp: any | never = null;
```
rickycao-qy, line 60: shall we use number as type instead of any

packages/cli/src/standalone-impl.ts
```diff
@@ -13,32 +15,90 @@ import {
...
+    let arr: Array<number | never> = [];
+    for ( i = 0; i < dataLength; i++ ) {
+      arr[ i ] = i;
+    }
+    for ( i = 0; i < dataLength; i++ ) {
+      index = Math.floor( Math.random() * (dataLength - i) ) + i;
+      if ( index !== i ) {
+        temp = arr[ i ];
+        arr[ i ] = arr[ index ];
+        arr[ index ] = temp;
```
rickycao-qy, line 70: `[arr[i], arr[index]]=[arr[index], arr[i]]` should work

packages/cli/src/standalone-impl.ts
Ahkari: okey

packages/cli/src/standalone-impl.ts
Ahkari: okey, its a better way for this situation.

packages/cli/src/standalone-impl.ts
```diff
@@ -13,32 +15,90 @@ import {
...
+      return [];
+    let i = 0;
+    let index = 0;
+    let temp: any | never = null;
+    let arr: Array<number | never> = [];
+    for ( i = 0; i < dataLength; i++ ) {
+      arr[ i ] = i;
+    }
+    for ( i = 0; i < dataLength; i++ ) {
+      index = Math.floor( Math.random() * (dataLength - i) ) + i;
```
WenheLI, line 66: Should use `seedrandom.js` instead of `Math.random()` for a more controllable random behavior.

packages/cli/src/standalone-impl.ts
```diff
@@ -13,32 +15,90 @@ import {
...
+    return Promise.all(buffer);
+  }
   seek(pos: number): Promise<void> {
     return this.dataAccessor.seek(pos);
   }
+  createRandom(dataLength = 0): Array<number> {
+    if (dataLength === 0)
+      return [];
+    let i = 0;
+    let index = 0;
```
WenheLI, line 59: `index` should be assigned in the below scope (the for loop one)

yorkie:
@FeelyChau Why was this PR closed without any reason?

FeelyChau:
@yorkie It's closed automatically, because the target branch has been deleted.

@Ahkari Could you recreate a PR to main branch, please?

PR #727, opened 2021-03-21
pipcook2.0: Support training data in a random way & with unit test cases

Ahkari:


FeelyChau (changes requested):

packages/core/src/utils/index.ts
```diff
@@ -91,3 +93,13 @@ export function shuffle(array: any[]): void {
...
+
+/**
+ * seedrandom
+ */
+export const seedrandom = seed;
+
+/**
+ * uuid
+ */
+export const uuid: typeof v4 = uuidv4;
```
FeelyChau, line 105: We already defined `generateId` for random id.

FeelyChau (changes requested):

packages/core/src/types/runtime.ts
```diff
@@ -90,23 +90,36 @@ export interface TableDataSourceMeta {
...
 
 export type DataSourceMeta = TableDataSourceMeta | ImageDataSourceMeta;
-export interface DataAccessor<T = DefaultType> {
+export interface SequentialDataAccessor<T = DefaultType> {
   next: () => Promise<Sample<T> | null>;
   nextBatch: (batchSize: number) => Promise<Array<Sample<T>> | null>;
   seek: (pos: number) => Promise<void>;
 }
+export interface DataAccessor<T = DefaultType> extends SequentialDataAccessor {
+  init: (size: number) => void;
```
FeelyChau, line 99: `init` should not be a part of interface `DataAccessor`, because script will never use it.

packages/core/src/types/runtime.ts
```diff
@@ -90,23 +90,36 @@ export interface TableDataSourceMeta {
...
+  nextRandom: () => Promise<Sample<T> | null>;
+  nextBatchRandom: (batchSize: number) => Promise<Array<Sample<T>> | null>;
+  resetRandom: (randomSeed?: string) => Promise<Array<number>>;
```
FeelyChau, line 102:
```suggestion
  resetRandom: (randomSeed?: string) => Promise<void>;
```
The script who call the `resetRandom` should not get  the result `Array<number>`.

FeelyChau (changes requested):

packages/core/src/types/runtime.ts
Ahkari: okey, remove `init` function definition from DataAccessor.

packages/core/src/types/runtime.ts
Ahkari: I want `resetRandom` has the same return  values with `createRandom` because developers may use values to do something like variables monitoring.

packages/cli/src/standalone-impl.ts
```diff
@@ -1,45 +1,105 @@
...
+      return null;
+    } else {
+      this.dataAccessor.seek(this.randomArray[this.randomIndex++]);
+      return this.dataAccessor.next();
+    }
+  }
+  nextBatchRandom(batchSize: number): Promise<Array<Sample<T>> | null> {
+    const buffer = [];
+    while (batchSize) {
+      buffer.push(this.nextRandom());
```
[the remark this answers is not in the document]
Ahkari: okey, I added if condition about it.

FeelyChau (dismissed):
LGTM.

WenheLI (dismissed):
LGTM

rickycao-qy (approved):
LGTM

yorkie (approved):

FeelyChau:
@yorkie @WenheLI @rickycao-qy Conflicts exist and need to be reviewed again.

yorkie (approved):

WenheLI (approved):
LGTM

rickycao-qy (approved):
LGTM

PR #730, opened 2021-03-23
cli: add command for script project initialization

FeelyChau:


yorkie (approved):

packages/cli/src/bin/pipcook.ts
```diff
@@ -3,12 +3,19 @@
...
+import * as fs from 'fs-extra';
 import { join } from 'path';
 import { constants } from '@pipcook/core';
 import { readJson, mkdirp, remove } from 'fs-extra';
 import { StandaloneRuntime } from '../runtime';
-import { logger, dateToString } from '../utils';
+import { logger, dateToString, execAsync } from '../utils';
+const download = require('download-git-repo');
 
+const templateMap: Record<string, any> = {
```
yorkie, line 14:
```suggestion
const templateMap: Record<string, string> = {
```

packages/cli/src/bin/pipcook.ts
```diff
@@ -3,12 +3,19 @@
...
 import * as chalk from 'chalk';
 import * as program from 'commander';
+import * as fs from 'fs-extra';
 import { join } from 'path';
 import { constants } from '@pipcook/core';
 import { readJson, mkdirp, remove } from 'fs-extra';
 import { StandaloneRuntime } from '../runtime';
-import { logger, dateToString } from '../utils';
+import { logger, dateToString, execAsync } from '../utils';
+const download = require('download-git-repo');
```
yorkie, line 12:
```suggestion
const download = import('download-git-repo');
```

packages/cli/src/bin/pipcook.ts
```diff
@@ -45,6 +52,39 @@ export const cacheClean = async (opts: CacheCleanOptions): Promise<void> => {
...
+  try {
+    const destDir = join(process.cwd(), name);
+    const pkgFile = join(destDir, 'package.json');
+    if (await fs.pathExists(destDir)) {
+      throw new TypeError(`${name} already exists`);
+    }
+    const [ type, branch = 'main' ] = scriptType.split('@');
+    const template = `${templateMap[type]}#${branch}`;
+    if (!template) {
+      throw new TypeError(`no template found for ${scriptType}, it should be one of 'datasource', 'dataflow', 'model'`);
```
yorkie, line 65: It's good to generate the string by `templateMap`.

packages/cli/src/bin/pipcook.ts
```diff
@@ -45,6 +52,39 @@ export const cacheClean = async (opts: CacheCleanOptions): Promise<void> => {
...
+    }
+    await new Promise<void>((resolve, reject) => {
+      download(template, destDir, { clone: true }, function (err: Error) {
```
yorkie, line 68: Why not using `utils.promisify`?

packages/cli/src/bin/pipcook.ts
```diff
@@ -45,6 +52,39 @@ export const cacheClean = async (opts: CacheCleanOptions): Promise<void> => {
   await Promise.all(futures);
 };
 
+export const createScriptRepo = async (scriptType: string, name: string): Promise<void> => {
```
yorkie, line 55: This function could be finished by `npm init`, see https://docs.npmjs.com/cli/v7/commands/npm-init/.

yorkie (changes requested):

packages/cli/src/bin/pipcook.ts
FeelyChau: `download-git-repo` has not types, so if `import` it, we will get error `ts(7016)`.

packages/cli/src/bin/pipcook.ts
FeelyChau: Does that mean we need to maintain another project named `create-pipcook-script`?

FeelyChau:
Add more details in the help description of command `pipcook script`, @yorkie could you review again, plz?

packages/cli/src/bin/pipcook.ts
yorkie: What we need is to rename our template to create-* projects.

packages/cli/src/bin/pipcook.ts
FeelyChau: Can you describe it more?

packages/cli/src/bin/pipcook.ts
yorkie: We need to leave a comment for this.

packages/cli/src/bin/pipcook.ts
```diff
@@ -3,11 +3,24 @@
...
-import { logger, dateToString } from '../utils';
+import { logger, dateToString, execAsync, downloadFromGit } from '../utils';
+
+const templateMap: Record<string, string> = {
+  'datasource': 'imgcook/pipcook-datasource-template',
+  'dataflow': 'imgcook/pipcook-dataflow-template',
+  'model': 'imgcook/pipcook-model-template'
+};
+
+const scriptTypeDescription = 'scriptType should be one of \'datasource<@branch>\', \'dataflow<@branch>\', \'model<@branch>\'.\n\
```
yorkie, line 19: Why not generating the string from `templateMap`?

packages/cli/src/bin/pipcook.ts
FeelyChau: If that, the create-* projects must be node packages first? Does it support branch?

FeelyChau:
We have a Pipcook script initializer now:
```sh
$ npm init pipcook-script datasource my-script -t ts
```
This PR should be closed now.

PR #738, opened 2021-03-29
core: move utils into cli, upgrade datacook to v0.0.5

FeelyChau:


packages/cli/src/standalone-impl.ts
```diff
@@ -1,124 +1,19 @@
...
-    this.randomIndex = 0;
-    this.randomArray = this.createRandom(this.size, seed);
-  }
-  next(): Promise<Sample<T> | null> {
-    return this.dataAccessor.next();
-  }
-  nextBatch(batchSize: number): Promise<Array<Sample<T>> | null> {
-    return this.dataAccessor.nextBatch(batchSize);
-  }
-  nextRandom(): Promise<Sample<T> | null> {
```
yorkie, line 39: Where do these methods go?

yorkie (changes requested):

packages/cli/src/standalone-impl.ts
FeelyChau: `DataCook.dataset.shuffle()` will instead of it.

yorkie (dismissed):

yorkie (approved):

WenheLI (approved):

PR #742, opened 2021-04-04
cli: fix file protocol uri with query

FeelyChau:
We should use `url.pathname` to get the real path so that support the query string.


yorkie (approved):

rickycao-qy (approved):
LGTM

PR #747, opened 2021-04-06
costa: support all entries in one script

FeelyChau:
Support all the entries exporting from one script. The script could export entries like:
```js
export = {
  datasource: datasourceEntry,
  dataflow: dataflowEntry,
  model: modelEntry
}
```


yorkie (approved):

yorkie:
Remember labeling issues and prs.

PR #749, opened 2021-04-06
cli: fix cli version error

FeelyChau:
Read from `package.json` at the root directory to make true we can get the correct version string.


rickycao-qy (approved):
LGTM

Issue #745, opened 2021-04-06
cli: feedback is required for every command


yorkie:
The commands `clean` and `install` don't give any feedback, we need to make sure every command will give a response.

FeelyChau:
@yorkie Should be fixed by #750.

yorkie:
Not yet, run `pipcook clean` doesn't get any feedback.

FeelyChau:
> Not yet, run `pipcook clean` doesn't give any feedback.

Not released. It will be published with `2.0.0-rc.3` tonight.

PR #750, opened 2021-04-06
cli: hidden command `install` and add logs for command `clean`

FeelyChau:
We should add the command `install` when it's ready.
Should fix issue #745 .


rickycao-qy (approved):
LGTM

PR #757, opened 2021-04-09
Feat/add resnet

WenheLI:


example/pipelines/resnet.v2.json
```diff
@@ -0,0 +1,15 @@
+{
+  "specVersion": "2.0",
+  "dataSource": "https://cdn.jsdelivr.net/gh/imgcook/pipcook-plugin-image-classification-collector@d00337c/build/script.js?url=http://ai-sample.oss-cn-hangzhou.aliyuncs.com/image_classification/datasets/imageclass-test.zip",
+  "dataflow": [
+    "https://cdn.jsdelivr.net/gh/imgcook/pipcook-plugin-process-tfjs-image-classification@db14a1a/build/script.js?size=224&size=224"
+  ],
+  "model": "https://cdn.jsdelivr.net/gh/imgcook/pipcook-plugin-tensorflow-resnet-model@1f680a6/build/script.js",
+  "options": {
+    "framework": "tf2@1.0.0",
```
yorkie, line 9: tf2's 1.0.0? It seems weird a bit :(

example/pipelines/resnet.v2.json
```diff
@@ -0,0 +1,15 @@
...
+    "train": {
+      "epochs": 1,
```
FeelyChau, line 11: As an example, we should set a normal value.

FeelyChau (changes requested):

yorkie:
Any updates? @WenheLI

WenheLI:
@yorkie - The only thing that needs to take care of is the naming of `framework`. `tf2@1.0.0` does sound weird. Any suggestions for the naming?

yorkie:
> @yorkie - The only thing that needs to take care of is the naming of `framework`. `tf2@1.0.0` does sound weird. Any suggestions for the naming?

tf@2.0.0?

WenheLI:
> > @yorkie - The only thing that needs to take care of is the naming of `framework`. `tf2@1.0.0` does sound weird. Any suggestions for the naming?
> 
> tf@2.0.0?

@FeelyChau - This might need cooperation.

FeelyChau:
@WenheLI @yorkie We can name it correctly if we can define the framework clearly. According to our current implementation, framework is a set of runtime dependencies of a pipeline, if we follow this definition, maybe `tf-resnet` will be a better name.

WenheLI:
> @WenheLI @yorkie We can name it correctly if we can define the framework clearly. According to our current implementation, framework is a set of runtime dependencies of a pipeline, if we follow this definition, maybe `tf-resnet` will be a better name.

IMO, the runtime that resnet depends on might have little to no differences with others. That means a specific runtime for `resnet` might be a bit weird

FeelyChau:
`resnet` has been implemented by #811 .

PR #761, opened 2021-04-11
fix warning from license and npm

FeelyChau:
Fix the warn as follows:
```sh
npm WARN lifecycle The node binary used for scripts is /var/folders/tq/cmmgdq_143jbd5w_drrdl8km0000gp/T/yarn--1618122759637-0.3013471152358005/node but npm is using /path/to/bin/node itself. Use the `--scripts-prepend-node-path` option to include the path for the node binary npm was executed with.
```
And license string.


WenheLI (approved):
LGTM

packages/cli/package.json
```diff
@@ -5,18 +5,18 @@
   "main": "dist/bin/pipcook.js",
   "scripts": {
     "test": "ava",
-    "build": "npm run clean && npm run compile",
+    "build": "yarn clean && yarn compile",
```
yorkie, line 8: yarn should not be required.

FeelyChau: Fixed.

WenheLI (dismissed):

rickycao-qy (approved):
LGTM

yorkie (approved):

PR #769, opened 2021-04-13
cli: import script without copy if the script protocol is `file:`

FeelyChau:
We could debug the scripts if the script files are not copied.


yorkie (approved):

rickycao-qy (approved):
LGTM

PR #776, opened 2021-04-16
cli: support relative path for script uri

FeelyChau:
As the [RFC-1808](https://tools.ietf.org/html/rfc1808#section-2.2), we could support the relative path for script uri like:
```json
{
    "datasource": "file:relative/path/to/script.js"
}
```
The real path is resolved as `${process.cwd()}/relative/path/to/script.js`.


yorkie (approved):

rickycao-qy (approved):
LGTM

PR #777, opened 2021-04-16
costa: make boa and datacook be global

FeelyChau:
The script is not adaptive for now.

### DIFF
--- a/packages/core/src/runtime.ts
+++ b/packages/core/src/runtime.ts
@@ -43,14 +43,6 @@ export type DataCookModule = typeof Datacook;
  */
 export interface ScriptContext {
   /**
-   * The boa module, the bridge between node and python.
-   */
-  boa: FrameworkModule;
-  /**
-   * DataCook module
-   */
-  dataCook: DataCookModule;
-  /**
    * This function can import a node module from framework. if the module not exists,
    * an error will be thrown.
    */
--- a/packages/costa/src/index.ts
+++ b/packages/costa/src/index.ts
@@ -11,13 +11,22 @@ import {
   PipcookFramework,
   ScriptType
 } from './types';
-import * as boa from '@pipcook/boa';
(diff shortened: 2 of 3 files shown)


packages/core/src/runtime.ts
```diff
@@ -42,14 +42,6 @@ export type DataCookModule = typeof Datacook;
...
-   * The boa module, the bridge between node and python.
-   */
-  boa: FrameworkModule;
-  /**
-   * DataCook module
-   */
-  dataCook: DataCookModule;
   /**
    * This function can import a node module from framework. if the module not exists,
    * an error will be thrown.
```
yorkie, line 47: importJS and importPY could be removed :)

yorkie:
Boa is improper as a global object because we don't expect it as a Web API.

packages/costa/src/index.ts
```diff
@@ -11,14 +11,25 @@ import {
...
 import Debug from 'debug';
 import { importFrom } from './utils';
 
 const debug = Debug('costa.runnable');
 
+declare global {
+  // eslint-disable-next-line no-var
+  var boa: typeof BOA;
+  // eslint-disable-next-line no-var
+  var datacook: typeof Datacook;
```
yorkie, line 26: Shall we make datacook to extend the global, thus user could use it just like:

```js
return Dataset.makeDataset({ trainData, testData }, meta);
```

yorkie:
That's great if developers still could import Python package via ESM:

```js
import * as np from 'py:numpy';
```

See https://github.com/imgcook/boa/blob/main/tests/es-module-loaders/py/test-esm-loader-custom.mjs.

yorkie:
Any updates @FeelyChau?

FeelyChau:
This PR could be closed, `boa` and `DataCook` could be required at runtime by excluding then when bundle.

PR #781, opened 2021-04-21
core: rename api for better semantic

FeelyChau:
Pipeline:
* dataSource -> datasource
* artifacts -> artifact
* specVersion -> version  ??

Runtime:
* dataSource -> dataset


docs/rfcs/0000-new-pipeline.md
```diff
@@ -15,7 +15,7 @@ Introducing a new Pipeline syntax for supporting:
 
 ```json
 {
-  "specVersion": "2.0",
+  "version": "2.0",
```
yorkie, line 18: The `version` is reversed for a pipeline itself.

yorkie:
And I'm -1 on renaming datasource to the dataset, the data source actually represents a data schema for multiple datasets.

docs/rfcs/0000-new-pipeline.md
FeelyChau: OK, rollbacked.

FeelyChau:
> And I'm -1 on renaming datasource to the dataset, the data source actually represents a data schema for multiple datasets.

If you mean the datasource type will be changed, I think we should confirm it before 2.0 releasing. Otherwise, if the interface type is exactly equal to `Dataset`, I think the name `dataset` is better, model script should not care the `dataflow`.

yorkie:
Dataset is the concept of datacook, the data source is Pipcook's.

FeelyChau:
> Dataset is the concept of datacook, the data source is Pipcook's.

What's the different except the name? In my opinion, they are the same.

yorkie:
Good question @FeelyChau, and I searched some for it:

- https://community.powerbi.com/t5/Service/Difference-between-data-source-and-data-set/m-p/531191
- https://www.sqlservercentral.com/forums/topic/what-is-the-relationship-between-datasources-and-datasets

In summary, they are not the same thing, and Data Source is a place where we fetch data or dataset, and the Dataset is a set of data, which represents the data itself. Therefore I'm agreed with using "dataset" :)

yorkie (approved):

PR #815, opened 2021-07-23
cli: add options `npmClient` and `registry` for artifact installing

FeelyChau:


packages/cli/src/bin/pipcook.ts
```diff
@@ -52,7 +56,7 @@ export const run = async (filename: string, opts: RunOptions): Promise<void> =>
     }
     pipelineConfig = await readJson(pipelinePath);
     // TODO(feely): check pipeline file
-    const runtime = new StandaloneRuntime(opts.output, pipelineConfig, opts.mirror, !opts.nocache);
+    const runtime = new StandaloneRuntime(opts.output, pipelineConfig, opts.mirror, !opts.nocache, opts.npmClient);
```
rickycao-qy, line 59: it seems that 'registry' is not passing to StandaloneRuntime constructor

FeelyChau: Fixed.

yorkie (approved):

PR #816, opened 2021-07-30
pipeline: support pipeline predict

FeelyChau:
* Move the dataset into `@pipcook/core` and rename it as `DatasetPool`.
* Add predictedData into `DatasetPool` for pipeline prediction.


packages/cli/package.json
```diff
@@ -28,26 +28,26 @@
...
-    "@types/request": "^2.48.5",
     "@types/semver": "^7.3.4",
     "@types/sinon": "^9.0.11",
     "ava": "^3.13.0",
     "import-fresh": "^3.3.0",
     "nyc": "^15.1.0",
     "sinon": "^10.0.0",
-    "ts-node": "^9.1.1"
+    "ts-node": "^9.1.1",
+    "typescript": "^4.3.5"
```
yorkie, line 50: Is the `typescript` required?

packages/cli/src/bin/pipcook.ts
```diff
@@ -22,6 +23,21 @@ export interface RunOptions {
...
+  // input for predict
+  str?: string;
+  uri?: string;
+  // Fetch the framework and script without cache
+  nocache?: boolean;
+  // Debug model
+  debug?: boolean;
+  mirror?: string;
+  // Development mode
+  dev?: boolean;
```
yorkie, line 40: Shall we avoid this kind of interface, which members are both optional :)

FeelyChau: One of `uri` and `str` should be passed, they should be optional and we could assert it.
The others have default value, fixed.

packages/cli/package.json
FeelyChau: I think each package should be self-contained, we need `typescript` to build ts script  to js script.

packages/cli/src/bin/pipcook.ts
```diff
@@ -87,15 +163,30 @@ export const cacheClean = async (): Promise<void> => {
...
     .option('--registry <registry>', 'npm registry for artifact installing')
+    .option('-d --debug', 'debug mode', false)
+    .option('--dev', 'development mode', false)
+    .option('--nocache', 'disabel cache for framework and scripts', false)
     .description('run pipeline with a json file.')
-    .action(run);
+    .action(train);
+
+  program
+    .command('predict <pipelineFile>')
```
yorkie, line 180: inference vs. predict? Which one should do we use?

packages/cli/src/constants/index.ts
```diff
@@ -35,3 +35,13 @@ export const PIPCOOK_FRAMEWORK_MIRROR_BASE = 'https://pipcook-cloud.oss-cn-hangz
  * framework description file name, which is located in the root directory of the framework package
  */
 export const FrameworkDescFilename = 'framework.json';
+
+/**
+ * framework description file name, which is located in the root directory of the framework package
```
yorkie, line 40:
```suggestion
 * framework description filename, which is located in the root directory of the framework package
```

yorkie (changes requested):
And the documentation should be updated :)

yorkie (dismissed):

yorkie (approved):

PR #858, opened 2021-09-06
costa: add framework initializer

FeelyChau:


yorkie:
The document is also required to be updated :)

FeelyChau:
@yorkie Updated.

yorkie (approved):

PR #865, opened 2021-09-17
pipeline: set backend for text classification

FeelyChau:
`tf.Unique` is not ready for `tfjs-node` and `tfjs-gpu`, so we need to specific the `backend` to `tfjs-cpu` for now.
see: https://github.com/tensorflow/tfjs/issues/4595


.github/workflows/build.yml
```diff
@@ -64,21 +64,22 @@ jobs:
       if: ${{ runner.os == 'Linux' && matrix.node_version == 12 }}
       run: |
         npm run cov && npm run cov:report
-    - name: Coveralls
-      if: ${{ runner.os == 'Linux' && matrix.node_version == 12 }}
-      uses: coverallsapp/github-action@master
-      with:
-        github-token: ${{ secrets.GITHUB_TOKEN }}
+    # Coveralls is down for maintenance, disable it for now
```
yorkie, line 67: We shouldn't disable it at the script, we could just merge the PR.

FeelyChau: The ci will fail If enable it, and we can't run the other checkings.

yorkie (approved):

PR #897, opened 2021-09-28
ci: upgrade node to v16

FeelyChau:


FeelyChau:
The build is failed on node v16 npm v7.10.0, it's a bug from npm, see https://github.com/npm/cli/issues/2632.
Maybe we have to wait for the new version of npm to fix it. The good news is that there is no problem running `pipbook`.

FeelyChau:
Walk around the bug from npm, the CI should be green now. @rickycao-qy

packages/cli/package.json
```diff
@@ -4,9 +4,10 @@
   "description": "",
   "main": "dist/bin/pipcook.js",
   "scripts": {
+    "preinstall": "(mkdir -p dist/bin && touch dist/bin/pipcook.js) || ((mkdir dist\\bin || echo 'exists') && echo '' > dist\\bin\\pipcook.js)",
```
yorkie, line 7: Do you have a try for https://github.com/charlesguse/run-script-os? This package can split the Windows/Linux script better :)

FeelyChau: But it does not work for `preinstall` script because the dependencies are not ready at this time.

rickycao-qy:
Is there a specific reason to upgrade node from 12 and 14 to 14 and 16

FeelyChau:
> Is there a specific reason to upgrade node from 12 and 14 to 14 and 16

Some users have started using node 16 and have encountered some problems. I think we should put 16 in CI to avoid this situation.

